PR #1, opened 2024-03-28
Change name to `Valkey`

AMoo-Miki:
### Description

This change replace references to placeholder names add corrects the repository destinations.

### Check List
- [X] Commits are signed per the DCO using `--signoff`

By submitting this pull request, I confirm that my contribution is made under the terms of the BSD-3-Clause License.

PR #3, opened 2024-04-01
removing cruft; switch CoC to match valkey-io/valkey

stockholmux:
### Description

- Removes some non-relevant template cruft
- Moves CoC from dummy text to Contributor Covenant to match valkey-io/valkey
 
### Issues Resolved

n/a

### Check List

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,23 +1,123 @@
-## Code of Conduct
+# Contributor Covenant Code of Conduct
 
-This code of conduct applies to all spaces provided by the Valkey project including in code, documentation, issue trackers, mailing lists, chat channels, wikis, blogs, social media and any other communication channels used by the project.
+## Our Pledge
 
-**Our open source communities endeavor to:**
+We as members, contributors, and leaders pledge to make participation in our
+community a harassment-free experience for everyone, regardless of age, body
+size, visible or invisible disability, ethnicity, sex characteristics, gender
+identity and expression, level of experience, education, socio-economic status,
+nationality, personal appearance, race, religion, or sexual identity
+and orientation.
+We pledge to act and interact in ways that contribute to an open, welcoming,
+diverse, inclusive, and healthy community.
 
-* Be Inclusive: We are committed to being a community where everyone can join and contribute. This means using inclusive and welcoming language.
-* Be Welcoming: We are committed to maintaining a safe space for everyone to be able to contribute.
-* Be Respectful: We are committed to encouraging differing viewpoints, accepting constructive criticism and work collaboratively towards decisions that help the project grow. Disrespectful and unacceptable behavior will not be tolerated.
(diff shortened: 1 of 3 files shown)


CODE_OF_CONDUCT.md
```diff
@@ -1,23 +1,123 @@
-## Code of Conduct
+# Contributor Covenant Code of Conduct
 
-This code of conduct applies to all spaces provided by the Valkey project including in code, documentation, issue trackers, mailing lists, chat channels, wikis, blogs, social media and any other communication channels used by the project.
+## Our Pledge
 
-**Our open source communities endeavor to:**
+We as members, contributors, and leaders pledge to make participation in our
```
madolson, line 5: Now is some type for stylistic questions I guess. We adopted one sentence per line here: https://github.com/valkey-io/valkey-doc?tab=readme-ov-file#styling-guidelines. Do you have thoughts about this?

stockholmux: @madolson I'm fine with that with one caveat: there are a few corner cases where it's not possible to achieve the desired rendered HTML with line breaks after every period (e.g. when using liquid templating to vary a sentence or in front-matter).

So, as long as we don't enforce this with some merge blocking linter, I'm good with it.

daniel-house: I see a "start each sentence on a new line" but nothing that says the sentence has to end on that line in https://github.com/valkey-io/valkey-doc?tab=readme-ov-file#styling-guidelines

One sentence per line can conflict with this other entry in the guidelines: "Please avoid writing lines that are too long, ..."

index.markdown
```diff
@@ -12,27 +12,16 @@ download_ctas:
 ctas:
     post: ' source on GitHub'
 
-headline: "Valkey makes it easy to find whatever you need in all of your data."
+headline: "Valkey: an open source, in-memory data store"
```
madolson, line 15: I think we're slightly aligned to move more away from in-memory.

```suggestion
headline: "Valkey: an open source datastructure store"
```

madolson (approved):
Sort of nitpick change we can cleanup later.

CODE_OF_CONDUCT.md
madolson: Yeah, I'm fully aligned with better defined style guides. The "one sentence per line" is something I've come to like specifically because of how it makes it clear when sentence structure is changing, but understand it's not always possible.

stockholmux:
Humm. Merging is blocked "The base branch requires all commits to be signed" but both commits are signed and DCO bot seems to be successful. 

Do we need to manually re-run the check?

index.markdown
daniel-house: Isn't in-memory the #1 way in which Valkey is being used?  Shouldn't we always be trying to ensure that none of the changes to support non-ram-storage will ever cause a degradation of in-ram performance?  It seems like a good idea to advertise our area of highest performance.

madolson: I don't disagree, but saying it's an "in-memory data store" short sells the other parts of it. Saying it's "primarily in-memory" also makes sense. I just think it's worth being opinionated about the differences.

PR #8, opened 2024-04-03
parsing the valkey command JSON files (draft only)

stockholmux:
### Description

Initial work to parse the [command json files](https://github.com/valkey-io/valkey/tree/unstable/src/commands) from Valkey. 

 - Add submodules for `valkey-io/valkey` and `valkey-io/valkey-doc`
 - Adds a bash script to generate stub files for all the commands
 - Adds recursive argument template
 - Adds a command layout (e.g. template for each command)
 - Two tweaks to SCSS to make it look readable and consistent
 - Adds a file to render the command listing
 - Adds all command stubs (generated by the bash script)

## Why so many files?

As you can see, **this is a massive amount of files**. 

To achieve the same URL structure as the previous site (e.g. `/commands/set/`), I need to have a single stub file (auto-generated by the bash file) that tells Jekyll to create and render the page from the data command JSON files. IMHO, this should probably not be committed to the repo, but rather generated on build and `.gitignored`. Undoing the last commit would re-enable that, if the current (and future) build infrastructure can support running the bash script on build.

## What's the status of the command parsing?

The command parsing in this PR is _acceptable but certainly not perfect_. I've cross checked it vs a few other sources and it looks good enough for someone to not be lead astray, but there are edges that don't look right. Given that there 400 commands, I've haven't personally checked everyone of them, but I tried to do a sampling check of similar commands to see how it looks it look wrong-ity wrong.

## Submodules

This PR uses submodules to valkey-doc and valkey. This is good as a stop-gap solution but I'm not certain this is the best long term solution. Like the node about the number of files, I think these should be pullled in on-build.

## What else is missing?

Gobs. 
- It doesn't currently handle ACL categories and RESP reply sections. 
- Similar sites seem to have more details on deprecation that don't appear to be present in the commands nor docs.
- The listing is bare-bones: it would be nice to have categories or some sort of better organization

But optimizing for good/fast not perfect, I think this PR is a decent place to start.

## What's next?

Working with `valkey-doc`, each command needs human eyes on it. I'd also like to figure out how we're handling non-Valkey versions in place. 

### Issues Resolved

n/a

### Check List


zuiderkwast:
I fail to see why we have to reinvent this wheel, when there is an existing way to do it that we can just clone.

Redict have their website up and running using Hugo: https://redict.io

stockholmux:
@zuiderkwast redict-io is `GNU Free Documentation License`, this repo is BSD-3 Clause.

zuiderkwast:
They cloned it from https://github.com/redis/redis-website apparently. That one is Apache 2.

Our website can be Apache 2 I guess? I can't see a reason we must stick to a BSD for the website. Do you?

stockholmux:
@zuiderkwast The best place to discuss the license of this repo is probably not in this draft PR.

Gemfile.lock
```diff
@@ -13,6 +13,7 @@ GEM
     eventmachine (1.2.7)
     ffi (1.16.3)
     forwardable-extended (2.6.0)
+    google-protobuf (3.25.3-arm64-darwin)
     google-protobuf (3.25.3-x86_64-darwin)
```
zuiderkwast, line 17: I don't know ruby gem deps... Do you pin one version for each architecture?

commands/acl-cat/index.html
```diff
@@ -0,0 +1,5 @@
+---
+layout: command
+title: acl-cat
+description: acl-cat.md
+---
```
zuiderkwast, line 5: Are these files generated? In that case, I don't think they should be checked in.

When all is in place, then when valkey-doc is updated, the website should preferably be updated automatically without any changes to the valky-website repo. This should include the addition of new commands, without adding files like this to the website repo.

_build/commands.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+# this bash script generates stubs directories/index files based on the command JSONs
+
+COMMANDS="./_data/commands/latest/*.json"
```
zuiderkwast, line 4: Where is the "/latest/" part coming from? From the valkey-doc repo?

stockholmux: `/latest/` is just where I set the internal representation (it doesn't affect the URLs). I wanted the `site.data.commands` namespace in Jekyll to be a object not a map, so I needed the subdirectory. It could be anything, but `latest` made sense.

commands/acl-cat/index.html
stockholmux: @zuiderkwast They are generated and I actually agree, the PR body `Why so many files?` and `Submodules` sections I talk about both points.

Right now, the build infrastructure for the site is a black box to me. Ideally, we'd bring these in with a GitHub action, but I think the site is built by some other means. 

But if you we merge this, it will work today without anything.

Gemfile.lock
stockholmux: that's actually a great question. I noticed it before and I had been manually removing the gem file from the commits, but it looks like I forgot on the last one. Obviously not a long-term solution. 

TBH, I'm not sure if it even matters if you have multiple different arch specific dependencies, but it's probably better to not.

I believe I need to add the architecture to the bundler [as described in this SO question](https://stackoverflow.com/questions/77738224/how-to-install-arm64-linux-based-ruby-gems-on-a-m1-arm64-darwin-based-mac). Let me double check if that works as expected.

commands/acl-cat/index.html
zuiderkwast: Right, I didn't read the description properly. :)

Yeah, let's merge this and we'll remove those files again in the future.

zuiderkwast (approved):

madolson (approved):

commands/acl-cat/index.html
madolson: I took a 10 minute pass to investigate this, and wasn't able to figure it out quickly. We nee to de-blackbox it and add documentation, but agree about not blocking on this.

madolson:
I missed the fact the top comment said draft only, so probably shouldn't have clicked merge.

stockholmux:
@madolson lol - that was my bad. What a comedy. 😓 

I switched it out of draft mode this afternoon and didn't change the PR. I think it's fine to merge (or remerge).

madolson:
As long as you don't mind, I looked through everything and the conversation and looked OK to me. I wanted to see it on the temp website.

stockholmux:
Hard to say, but looks like the submodules didn't get pulled in to the preview site. I can see the [commands page](https://d18m7x5q574jh8.cloudfront.net/commands/) (which should be a list of commands) and the [individual commands pages](https://d18m7x5q574jh8.cloudfront.net/commands/hset/), but they don't have anything on them.

They should look something like this:
Screenshot 2024-04-07 at 9 42 22 PM
Screenshot 2024-04-07 at 9 42 08 PM


@AMoo-Miki Can you take a look workflow that is building the preview site?

PR #17, opened 2024-04-08
Use GitHub Pages for hosting

AMoo-Miki:
### Description

This change will build and deploy the website and docs to GHP. This should be followed by renaming the repository to `valkey-io.github.io`.
 

### Check List


madolson:
The OpenSearch once commented this was a bad idea because it made later redirection more difficult. I assume we'll still be able to have URLs like valkey.io/docs/commands/set  or will it have to be valkey.github.io/whatever ?

Laphatize:
Assuming path structure is the same in both domains, there should be no issue.

The below is possible with Github Pages:
https://valkey.github.io/docs/commands/set  -> https://valkey.io/docs/commands/set



More info regarding custom domains & Github pages:
https://docs.github.com/en/pages/configuring-a-custom-domain-for-your-github-pages-site/managing-a-custom-domain-for-your-github-pages-site

madolson (dismissed):
Nothing strikes me, but I'm less familiar about how all of this works.

AMoo-Miki:
The only complication arises if we want more control or want to do things beyond the super limited ones GHP allows us to. For example, OpenSearch wanted to have different versions of their docs to be stored in different branches and built only when they had changes (because of the time cost due to its volume). GHP is okay for what we have today but if our needs evolve, we can easily move to something with more freedom.

We have the plumbing for both GHP and CloudFront solutions. While access to the plumbing on GHP can be delegated to anyone, I have no way to share access to the CloudFront plumbing (which is my main motivator for this PR).

.github/workflows/deply.yml
```diff
@@ -0,0 +1,55 @@
+name: Deploy Jekyll site to Pages
+
+on:
+  push:
+    branches: ["main"]
```
stockholmux, line 5: can we make this a "production" branch?

stockholmux (changes requested):
One small change - I'd like to have a `main` branch and a `production` branch. That we can merge and integrate change on main and protect production as needed.

stockholmux:
@madolson The redirect strategy (although limited) is fine as long as we don't have people depending on `valkey-io.github.io` and the final destination is an actual URL (valkey.io). 

(you probably heard me preaching that `.github.io` is a bad move. GitHub Pages is fine behind a domain name though)

stockholmux:
Btw, great work on this @AMoo-Miki!

madolson:
> (you probably heard me preaching that .github.io is a bad move. GitHub Pages is fine behind a domain name though)

Sounds good. We are writing valkey.io everywhere, I just want to make sure is no concerns with using that.

.github/workflows/deply.yml
madolson: Minor preference for "prod", since that is what I'm used to. Then again I also prefer mainline over main, so it might just be my preference for "Amazon" stuff.

AMoo-Miki: Switched to `prod`.

madolson (approved):

stockholmux (approved):
LGTM. Ship it.

PR #22, opened 2024-04-09
Add the ability to pull in docs from `valkey-doc`

stockholmux:
### Description

- Adds doc import layout (see below)
- Adds example of using doc import layout for `docs/management/persistence.md`
- Adds another symbolic link to the `_includes` folder for the markdown docs from `valkey-doc` in the workflow
- Updates submodule

#### What does the `doc-import` layout do?

This allows the consumption of the existing markdown files in `valkey-doc` by stripping out the existing frontmatter (which is not relevant). Alternately, if there isn't front matter on the file (it doesn't start with `---`) it still works.

This way we can keep the concerns separated: `valkey-doc` contains content, `valkey-website` contains metadata and site specific (templates, forward, etc.) information. 
 
### Issues Resolved

n/a

### Check List


madolson (approved):
Generally lgtm

PR #31, opened 2024-04-11
Logo & Favicons

chris-short:
- Added multiple formats of the logo to /img
- Added full compliment for favicons
- head.html edited accordingly

### Check List


assets/img/favicon/manifest.json
```diff
@@ -0,0 +1,41 @@
+{
+ "name": "App",
+ "icons": [
+  {
+   "src": "\/android-icon-36x36.png",
```
stockholmux, line 5: the `\/andorid....` looks weird. Should it be escaped like that?

chris-short: I've seen it both ways. But, these paths aren't accurate either. I'll need to edit this regardless.

chris-short: Pushed an update

Dockerfile
```diff
@@ -0,0 +1,35 @@
+# Build the Ruby image from source
+FROM ruby:3.0 AS ruby-build
```
stockholmux, line 2: Can you contribute the Dockerfile under a different PR?

stockholmux (changes requested):
Favicons seem to be working, but the Dockerfile should go elsewhere

Dockerfile
chris-short: That wasn't supposed to be here...  
I loathe git

chris-short: Removed from this PR.

stockholmux (approved):
LGTM

PR #67, opened 2024-05-11
Moves from Jekyll to Zola

### LINKED ISSUE #62: Move to Zola
Several folks in other issues questioned the use of Jekyll, but in the rush to create an initial website it was used out of convenience and at hand expertise to get it out quickly.

Now with a little breathing room, I think it's time to evaluate moving to a new static site generator (SSG). SSGs are very sticky after your site grows and this will only get harder for this site as time goes on. Investing in Jekyll right now seems like an issue on a few fronts:

1) Jekyll is Ruby and requires some tricky setup for those who don't have Ruby already setup on their machine: you have to get the right Ruby version, install the package manager, etc. There is a container but even that is slightly tricky to use and troubleshoot.
2) Even in the incomplete state the rebuild time is pretty annoying. On my machine it takes about 15 seconds to rebuild and that will only get worse in time.
3) Jekyll has some specific requirements for paths to include files. The project inherited a somewhat complicated structure: docs live in one repo (`valkey-io/valkey-doc`), the command JSON source in another (`valkey-io/valkey`) and the website in a third, in Jekyll, this lends itself to using git submodules (which are their own problem).
4) The inherited structure and some choices made in `valkey-doc` requires a decently powerful templating language. Liquid can do this but it's a little messy to achieve what we need to achieve for this site. 

I've investigated a few different SSGs recently and tried to see if they can easily handle our site and will be a good choice for years to come. 

I built a few proofs of concept and the one that comes out on top based on my evaluation is [Zola](https://www.getzola.org/). 

1. It's an easy-to-install single binary, built in Rust.
2. It's much, much, much quicker: the site renders in 100-150ms  vs 15 seconds on Jekyll (~150x faster)
3. The directory structure is flexible enough to do away with submodules,
4. Zola uses [Tera](https://keats.github.io/tera/docs/) seems up to the job. Tera is fairly close to Liquid, but a little more strongly typed and has explicitness which will be useful going forward.

Here is the plan, unless someone majorly objects:

1. Create a branch for the zola version.
2. Move content over to Zola. 
3. Continue to operate the Jekyll site, adding new **content only** to Jekyll.
4. The zola version will focus on getting the template and existing content to build locally and match the Jekyll side.
5. Once zola seems feature complete for template, freeze content on the Jekyll side and migrate anything new to Zola
6. Swap around `prod` and `main` branch to use Zola.

Move forward with the new world.

stockholmux:
### Description

This changes valkey.io from Jekyll to Zola and adds all documentation topic pages.

**Sorry for the big bang** there isn't a graceful way to do this in git.

#### Path forward to Zola in production

1. Cut a branch for the existing Jekyll site. We should be able to still merge this new branch to prod. (✅ [`jekyll-version`](https://github.com/valkey-io/valkey-io.github.io/tree/jekyll-version)) 
1. Merge to this PR to main.
1. Assign reviewers for the documentation topics. I'd like a set of eyes to confirm that each topic works in the template on a local build and isn't completely wrong before we go to full production. 
   -  If the content is super wrong or problematic, there is now a way to flag the topic and display a message atop the page that says "**This page is under review.** The page is likely incorrect, contains invalid links, and/or needs technical review. In the future it may change substantially or be removed entirely." by adding it to the `config.extra.review_list` array.
   - Flagged content doesn't block publishing the site, but a issue on `valkey-io/valkey-doc` should be created at the same to address the content bits.
1. Setup a new GitHub Action to publish the Zola site (see [Zola docs](https://www.getzola.org/documentation/deployment/github-pages/#github-actions)). It will need some custom for the specific build scripts, but nothing crazy.
1. Any new content merged to the Jekyll branch is ported to Zola (should be minimal).
1. Publish initial production Zola build
1. Delete Jekyll branch for good.

 
### Issues Resolved

Fixes #62

### Check List


madolson:
Do you have a suggested way to review this?

README.md
```diff
@@ -13,21 +13,49 @@ If you discover potential security issues, see the reporting instructions on our
 
 ## Build Locally
 
-This site is built using Jekyll 4.3.2 with Ruby 3.3.0. Other versions may work, but YMMV.
+This site is built with [Zola](https://www.getzola.org/).
```
madolson, line 16: I can't comment on the line I want, but there is a mention Git submodules at the top of the Readme which is a lie now.

README.md
```diff
@@ -13,21 +13,49 @@ If you discover potential security issues, see the reporting instructions on our
...
+All files created in this process are ignored by git.
+Commit your changes to your local copy of `valkey-io/valkey-doc`.
+
+### Building the command reference
+
+Let's say that this repo and your local copy of `valkey-io/valkey-doc` and `valkey-io/valkey` reside in the same directories.
+From the root directory of this repo run:
+
+```shell
+./build/init-commands.sh ../valkey-doc/topics ../valkey/src/commands
```
madolson, line 52: I don't read ahead, so I ran this command `./build/init-commands.sh` based off a comment from the base command page when I saw it. If you run this, it still tries and fails to create the symbolic links and they are broken. It prevents future attempts to run the commands. Probably worth an issue, but we should probably fix them so we validate something is there before creating bad symlinks.

madolson: Also, isn't this wrong? I think this should be:
```suggestion
./build/init-commands.sh ../valkey-doc/commands ../valkey/src/commands
```

madolson:
When running it locally, I don't get the valkey logo in the title like valkey.io currently:
image. I just see Valkey.

templates/commands.html
```diff
@@ -0,0 +1,36 @@
+{% extends "fullwidth.html" %}
+
+{% import "macros/command.html" as commands %}
+
+{% block main_content %}
+{% if section.pages | length == 0 %}
+<strong>No command pages found.</strong> Did you run <code>./build/init-commands.sh</code>?
```
madolson, line 7: Let's point to the README.md here instead, since this command doesn't work.

madolson:
Markdown has looks a bit weird:

image

compared to 
image

Not sure what the long term is, but it looked a bit better before.

madolson:
Not a major issue, but it looks like we aren't rendering images correctly from docs/topics. http://127.0.0.1:1111/docs/topics/lru_comparison.png just gives a generic error.

madolson:
Ok, I clicked through all the links I could find and everything else looks good to me.

README.md
stockholmux: Yeah, you're right about the paths.

I've added both argument presence validation as well as a check to see if the paths exist.

stockholmux:
@madolson I figured out the poor syntax highlighting. Looks like I was [missing some recommended styles](https://www.getzola.org/documentation/content/syntax-highlighting/#styling-codeblocks) for the built-in syntax highlighting scheme.

Screenshot 2024-05-16 at 8 05 23 AM

This also solves some weird container overflow problems. It still uses the light-on-dark (which I prefer for code blocks), but I think it's possible to make it dark-on-light if you feel strongly.

madolson:
@stockholmux That looks fine to me, the problem was the lack of padding and the weird overflowing you observed. The proposal you have LGTM.

content/authors/dmitrypol.md
```diff
@@ -1,8 +1,8 @@
 ---
-name: Dmitry Polyakovsky
-short_name: dmitrypol
-photo: '/assets/media/authors/dmitrypol.jpeg'
-github: dmitrypol
+title: Dmitry Polyakovsky
+extra:
+    photo: '/assets/media/authors/dmitrypol.jpeg'
+    github: dmitrypol
```
madolson, line 5: This isn't rendering correctly, it's just showing the text and not the photos.

madolson (approved):
This mostly LGTM, GitHub isn't handling this massive PR very well and would rather merge it and work incrementally on smaller diffs.

madolson:
Created an issue documenting all of the files that still need to be reviewed: https://github.com/valkey-io/valkey-doc/issues/91

stockholmux:
@madolson when you say "would rather merge it and work incrementally on smaller diffs" are you asking me to abandon this PR and do it in multiple PRs?

content/authors/dmitrypol.md
stockholmux: oh I see. It should show a posts by list(https://valkey.io/authors/kyledvs/), but this is a relatively easy fix.

madolson:
No I'm saying merge it!

PR #80, opened 2024-05-31
Deployment GitHub action for zola version

### LINKED ISSUE #11: Jekyll is good idea as static website builder ?
Jekyll is pretty cool to use and easy to deploy. But if you want to scale a project with component such as dropdown it's more complicated.
Also for i18N its isn't easy to maintain.

it's open question.

stockholmux:
### Description

This is a github action that builds the zola version of the website. This is a stepping stone to the deployment script.

### Issues Resolved

#11 

### Check List

### DIFF
--- a/.github/workflows/zola-deploy.yml
+++ b/.github/workflows/zola-deploy.yml
@@ -33,7 +33,12 @@ jobs:
           ./build/init-topics.sh ../valkey-doc/topics 
           ./build/init-commands.sh ../valkey-doc/commands ../valkey/src/commands
 
-      - name: Build and deploy
+#      - name: Build and deploy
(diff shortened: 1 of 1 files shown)


chris-short:
FWIW, this seems sane to me. You're controlling the process mostly through established scripts so as long as those run right, you should be good here. Have you tested this on a dummy copy of this repo? If it works and the logs aren't giving you any red flags, I think you're good.

stockholmux:
Did some additional testing based on @chris-short's advice and confirmed that this builds and creates a correct artifact. While I couldn't test it through to the github pages deployment stage, it is the same method used in the Jekyll deployment.

PR #95, opened 2024-07-05
Push to prod: relocate `/docs/topics` to `/topics`

stockholmux:
This publishes the topics at `/topics` instead of `/docs/topics`


madolson (approved):
Checked out, seems to work fine.

PR #111, opened 2024-08-16
Adding analytics to valkey.io

chris-short:
### Description

Adding Fathom Analytics to site

Apparently I was supposed to do this when the site was first stood up but, as an unfunded project, LF didn't mention anything about it and I was more focused on getting a working site up.

Analytics are important as we make changes and add content to the site. The data will enable us to find what pages perform well vs. those that don't. It can also enable discovery of how users use the site itself, which docs are more important to users, and a bevy of other useful data as the project grows.
 
### Issues Resolved

N/A

### Check List

### DIFF
--- a/templates/includes/head.html
+++ b/templates/includes/head.html
@@ -39,9 +39,9 @@
     <link rel="alternate" type="application/atom+xml" title="Atom feed" href="{{ get_url(path="atom.xml", trailing_slash=false) }}" />
 
 
-<!-- Fathom - beautiful, simple website analytics -->
-<script src="https://cdn.usefathom.com/script.js" data-site="FRZJYDAH" defer></script>
-<!-- / Fathom -->
+    <!-- Fathom - beautiful, simple website analytics -->
+    <script src="https://cdn.usefathom.com/script.js" data-site="FRZJYDAH" defer></script>
+    <!-- / Fathom -->
 
     {% if page and page.extra and page.extra.head_extra %}{{ page.extra.head_extra }}{% endif %}
 </head>
--- a/templates/includes/head.html
+++ b/templates/includes/head.html
@@ -39,15 +39,15 @@
     <link rel="alternate" type="application/atom+xml" title="Atom feed" href="{{ get_url(path="atom.xml", trailing_slash=false) }}" />
 
 
-    <!-- Google tag (gtag.js)
-    <script async src="https://www.googletagmanager.com/gtag/js?id=G-6M0FF13NRB"></script>
+    <!-- Google tag (gtag.js) -->
+    <script async src="https://www.googletagmanager.com/gtag/js?id=G-L8FMHY9BG0"></script>
     <script>
       window.dataLayer = window.dataLayer || [];
       function gtag(){dataLayer.push(arguments);}
       gtag('js', new Date());
 
-      gtag('config', 'G-6M0FF13NRB');
-    </script> -->
+      gtag('config', 'G-L8FMHY9BG0');
+    </script>
 
     {% if page and page.extra and page.extra.head_extra %}{{ page.extra.head_extra }}{% endif %}
 </head>
--- a/templates/includes/head.html
+++ b/templates/includes/head.html
@@ -39,15 +39,9 @@
     <link rel="alternate" type="application/atom+xml" title="Atom feed" href="{{ get_url(path="atom.xml", trailing_slash=false) }}" />
 
 
-    <!-- Google tag (gtag.js) -->
-    <script async src="https://www.googletagmanager.com/gtag/js?id=G-L8FMHY9BG0"></script>
-    <script>
-      window.dataLayer = window.dataLayer || [];
-      function gtag(){dataLayer.push(arguments);}
-      gtag('js', new Date());
-
-      gtag('config', 'G-L8FMHY9BG0');
-    </script>
+<!-- Fathom - beautiful, simple website analytics -->
+<script src="https://cdn.usefathom.com/script.js" data-site="FRZJYDAH" defer></script>
+<!-- / Fathom -->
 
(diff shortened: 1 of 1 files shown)


madolson:
Can you add more of a description here of what we are doing. What data are we exactly tracking from users? Are we just tracking per-page access statistics? I understand there is a desire to collect more data, but I want to weigh that against the fact it will worsen the performance of the site and I don't like that everywhere I go on the internet I'm being tracked.

zuiderkwast:
Please don't. We'd need to add annoying cookie warning banners. Besides, I'm personally against that kind of user tracking.

katiemcollins:
We need to measure how many people we’re reaching with this content. To raise awareness and grow adoption, it is essential to have data that we can see # of viewers, etc to verify if these methods are working and getting to people.

chris-short:
There are privacy respecting analytics tools out there (Fathom & Plausible). They are both paid options but, do offer the cookieless experience I believe you're after.

I use Fathom personally and both tools are currently up on my website [chrisshort.net](https://chrisshort.net/) for your inspection. If you need a cookie testing tool, this one appears to work well: https://www.cookieyes.com/cookie-checker/  
  
I'm hesitant to go about setting up accounts for either of them as demos because of two reasons:  
  
1\. They're paid and we'd have to figure out the most sustainable way to manage payment going forward (not sure if there's a process already for paying for things under the Valkey banner)  
2\. If/when Valkey becomes a funded project, LF will take control of the web site and that analytics data would be lost as LF would add Google Analytics almost immediately.

The import options into Google Analytics is limited based on my experience and the docs don't give me any hope its improved. If we're okay with potentially losing the site's analytics data history then it's worth exploring Fathom and Plausible. If you'd like to keep that data in a future where Valkey becomes a funded project, then bringing Google Analytics (and a lightweight analytics cookie notice) is the answer.  
  
I will say, without analytics I'm incredibly hesitant to do more work on the web site as I would not be able to discern if any changes made help or hurt the site/project overall.

zuiderkwast:
Statistics can be found in old-fascioned HTTP server logs, but I suppose don't have access to that from github pages. We could if we host it on another webserver though. (Then we could also get 301 redirects. That's another think we're missing from github pages.)

@chris-short the cookieless ones you mentioned would be my preference. I'm not sure if we'll ever move to a paid LF project. We seem to prefer that Valkey remains a LF community project.

I will not put in my veto on this though.

chris-short:
A web server log means we'd be going from a globally load balanced, CDN backed service to a single server (and maintaining infra to pull the logs out to a team accessible place). This is not advisable in modern times as I've seen this fail repeatedly plus the people asking me for analytics wouldn't understand an HTTP log file. I would be adamantly against such a step back in time.  
  
If redirects are a goal, and Zola/GH Pages don't support them (they don't appear to, which I find really odd), the hosting providers I've used in the past (Netlify & Vercel) have configurable redirects in their respective config files. At this point, earlier in the website's life, I would consider re-platforming to something more suited to those needs (Hugo & Netlify is the stack I'm most experienced with from other OSS projects). I would strongly advise not standing up separate infra to handle redirects unless there's just so many we have to (think 1000s of redirects).

zuiderkwast:
Yeah, we can live without redirects. I realize I'm living in the past (a decade or two) when it comes to web infra. :)

Anyway, regarding annoying cookie banners for non-essential cookies: Don't we need to have those to comply with EU and US law? That in itself is a reason to avoid tracking cookies IMO.

The purpose of those laws was to encourage websites to stop using non-essential 3rd party tracking cookies, but that strategy didn't seem to work. It seems that only websites like wikipedia.org, duckduckgo.com, gnu.org, debian.org and such sites care about this stuff.

Cookie banners are not necessary for essential cookies, like for a login session, since those serve a purpose to the user. I remember GitHub remove their tracking cookies and the cookie banners some years ago (see https://techcrunch.com/2020/12/17/github-says-goodbye-to-cookie-banners/ and https://github.blog/news-insights/company-news/no-cookie-for-you/) but the latter now displays a cookie banner anyway, after a recent update...

chris-short:
So the law behind cookie banners didn't work as intended like you mentioned. The cookie notices are not required for essential functions (even dark mode/light mode cookies are exempt is my understanding). Using cookieless analytics solutions also does not require a cookie banner because there are no cookies to put up a banner about (because law and intent are two very separate things).

chris-short:
Goog removed.
Fathom analytics added.
Had to rebase to signoff latest commit.

zuiderkwast (approved):
Great! Is everyone OK with this?

chris-short:
When this merges, stats will be available here: https://app.usefathom.com/share/frzjydah/valkey.io

templates/includes/head.html
```diff
@@ -39,15 +39,9 @@
...
-    <script>
-      window.dataLayer = window.dataLayer || [];
-      function gtag(){dataLayer.push(arguments);}
-      gtag('js', new Date());
-
-      gtag('config', 'G-6M0FF13NRB');
-    </script> -->
+<!-- Fathom - beautiful, simple website analytics -->
+<script src="https://cdn.usefathom.com/script.js" data-site="FRZJYDAH" defer></script>
+<!-- / Fathom -->
```
madolson, line 44:
```suggestion
    <!-- Fathom - beautiful, simple website analytics -->
    <script src="https://cdn.usefathom.com/script.js" data-site="FRZJYDAH" defer></script>
    <!-- / Fathom -->
```

madolson (approved):
Made a nitpick. I read through the documentation and I feel a lot better about using Fathom.

hpatro:
> https://app.usefathom.com/share/frzjydah/valkey.io

Does this require a new deployment @madolson @stockholmux ? The dashboard (last week) has no data to display at least for me.

PR #119, opened 2024-09-01
Some suggestion for improving the blog

madolson:
### Description

While writing the Valkey 8.0 blogs, I had some minor gripes about the formatting and thought I would try to present them in a PR and get some feedback.

1. On desktop, the real estate dedicated to authors is too much. I prefer the style of having authors at the end, so moved them down and aligned it horizontally.
2. The previous/next links are squished together. So now the right link is right adjusted. 
 
What it looks like now:

image

### Issues Resolved

No issues, just thought I would throw it out here as a suggestion.

### Check List


PingXie (approved):
fully agreed on both points. I don't know anything about .scss files so approving the high level ideas only.

sass/_valkey.scss
```diff
@@ -259,7 +259,7 @@ p {
   height: 100%;
   margin: 0 auto;
 
-  .container.right-aside img {
+  .container img {
```
stockholmux, line 262: `.container img` is pretty broad. Every template uses `.container`, so I'd ideally would prefer it to be a more specific class.

That said, I don't *think* currently we have images that doesn't need to span the container and (eventually) we probably need to work on a better template making this all more/less obsolete, so I think it's fine for now.

templates/blog-page.html
```diff
@@ -15,24 +15,24 @@ <h1>{{ page.title }}</h1>
...
+</div>
 <hr />
 {% include "includes/share-buttons.html" %}
 <div class="blog-nav">
   {% if page.higher and page.higher.title %}
   <a class="prev" href="{{page.higher.permalink}}">&laquo; {{page.higher.title}}</a>
   {% endif %}
   {% if page.lower and page.lower.title %}
-  <a class="next" href="{{page.lower.permalink}}">{{page.lower.title}} &raquo; </a>
+  <a style="float:right;" class="next" href="{{page.lower.permalink}}">{{page.lower.title}} &raquo; </a>
```
stockholmux, line 35: Inline styles are gauche, hard to maintain, and you can save a few bytes with cached CSS. 

For the same effect, you can revert it to the previous, and add `_valkey.scss`
```css
.next {
  float: right;
}
```

stockholmux (changes requested):
I put a few comments on the code itself.

As far as the idea, I can get behind this if this is what lots of folks want it, but the shorter lines were probably easier to read and more accessible.

The [WCAG 2.1](https://www.w3.org/WAI/WCAG21/Understanding/) states that there should be no more than 80 characters per line.


> For people with some reading or vision disabilities, long lines of text can become a significant barrier. They have trouble keeping their place and following the flow of text. Having a narrow block of text makes it easier for them to continue on to the next line in a block. Lines should not exceed 80 characters or glyphs (40 if CJK), where glyphs are the element of writing in the writing system for the text.


After the change, lines are 140-155 characters per line. Even before the change, it was often over 90. I will say that the site is probably far from meeting the WCAG 2.1 in many other areas, but I do think it's important to go towards these standards.

madolson:
> As far as the idea, I can get behind this if this is what lots of folks want it, but the shorter lines were probably easier to read and more accessible.

Hmm, I don't like the short lines, but you did present other data and now I don't know what to say. I think my main grep was that for code lines, like the ones in the bitnami blog, are pretty annoying to read with 80 lines. Generally I don't mind the 80 line length though. It does say "should", so it feels like we are allowed to ignore it.

madolson:
Honestly, the 80 character isn't really a hill I want to die on, and it seems like the docs and commands use it. I really just didn't like the clutter with the authors at the top. The latest version keeps the width and is much cleaner.

madolson:
image
The next/prev is still really close, not sure there is much to do about that without putting them on different lines.

stockholmux:
> The next/prev is still really close, not sure there is much to do about that without putting them on different lines.

I agree, that's messy.

What if we did something like this (of course with the writer bios at the bottom).

Screenshot 2024-09-04 at 8 44 51 AM

I was able to mock it up in dev tools by altering the styles for next/prev to something like this:
```css
    display: inline-block;
    float: right;
    max-width: 40%;
    text-align: right;
```

(the other side is slightly different, but you can see the approach)

It gets complicated because we need to undo this on mobile or it will be like 3 characters per line on the next/prev. I'm glad to do this, but maybe we should make an issue and I can just implement it under a different PR.

stockholmux (commented):
LGTM

PR #121, opened 2024-09-03
Adds the ability to add a top banner

stockholmux:
### Description

- Adds a top banner to the site on every page that is configurable by changing the `config.toml` file. If the configuration item is absent, the banner disappears
- Adds a message about Valkey Developer Day (to be removed the day before the event)

### Issues Resolved
n/a

### Check List

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -57,3 +57,5 @@ publish_hold = [
     "/topics/protocol/",
     "/topics/rdd/"
 ]
+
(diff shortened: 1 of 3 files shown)


madolson (approved):
Sure. Feels a bit late to add it now, but it's useful to have moving forward. Is there any way we could automatically configure the banners with dates so they get removed automatically?

PR #122, opened 2024-09-04
Fixes command topics with underscores not showing on command referenc…

### LINKED ISSUE #117: [BUG] Commands with underscore in the name missing on website
**Describe the bug**

The following commands contain an underscore and they're not displayed on the website:

```
BITFIELD_RO
EVAL_RO
EVALSHA_RO
FCALL_RO
GEORADIUSBYMEMBER_RO
GEORADIUS_RO
SORT_RO
```

kris-ms:
…e page

### Description

The Zola slugifier disintegrates generated page slugs containing underscores when left to its own devices.
This also means that we can not fetch JSON data from the Valkey repo as we use these slugs as arguments to the JSON loader macro.

A small config change corrects this while still removing unsafe characters from the generated slugs. The topics now appear and fetch data appropriately.
 
### Issues Resolved

Closes #117 

### Check List


stockholmux:
@kris-ms I'm trying to understand the specific cases which this is causing problems - could you ELI5?

I built this version vs the `main` and diff'ed the tree and I couldn't see an actual difference in the files generated, so I feel like I'm missing something.

kris-ms:
Sure, its a slippery one as it won't affect the file tree at all, it is an issue with Zola itself relating to it parsing slugs from filenames in `/content/`.

 The easiest place to spot the error would be to look at https://valkey.io/commands/ and scroll down to `SORT`.

image

Note that `SORT_RO` does not exist here, which it should.
It appears that Zola is aggressive with its slugification, and is trimming the filename suffix as it contains a "_" character. The affected commands are:

```
BITFIELD_RO
EVAL_RO
EVALSHA_RO
FCALL_RO
GEORADIUSBYMEMBER_RO
GEORADIUS_RO
SORT_RO
```

If we check out line 11 in `/templates/commands.html` we see that we actually rely on these slugs to pull JSON data from the Valkey repo:

```rust
{% set command_data = load_data(path= commands::command_json_path(slug= page.slug), required= false) %}
```

Which means that we are never pulling any data or generating pages for any of the above `*_RO` commands.

With this quick config change, the slugification process is eased up a little bit to allow for underscores in page slugs, choosing "safe" means that we still remove unsafe characters from the filenames and don't have broken paths.

Fixed version running locally: 

image

Super long ELI5 but I hope that makes sense!

stockholmux:
@kris-ms Gotcha! Great catch, actually. I was really trying to square what it was doing, I thought sure it was the slugification on the file creation end, so thanks for the clarification.

stockholmux (approved):

PR #143, opened 2024-09-27
Add branding guidelines page

### LINKED ISSUE #107: Create a branding page for Valkey
We have our Valkey branding assets located here, https://github.com/valkey-io/assets, but ideally there should be a nice web page also showing the same information. Ideally we should at least link to the branding.

kris-ms:
### Description

With the recent creation of the branding guidelines for Valkey, a web page with the information is in order.
This PR creates a draft for the branding page and is a spot for discussion on improvements/fixes to said page.

I'm a terrible writer so I've not made any real attempts at prose so if there are any suggestions I'm happy to implement!

A screenshot for ease of checking:

image
 
### Issues Resolved

Closes #107 (maybe)

### Check List


stockholmux:
@kris-ms I'm in favour of adding better web presence for the brand guidelines but I think I'd like them to live in one place, not having it both as a PDF and on the website. I think this PR duplicates everything from the PDF, which is a potential for things to go out-of-sync and causing confusion.

FWIW: The current brand guidelines are good but not quite complete (e.g. it doesn't have things like fonts). So, I do expect them to change. 

The assets may have value to have in the repo vs linked to GitHub, but we can do some GitHub Action magic to pull them in without committing them to the repo directly (avoiding duplication and potential out-of-date assets).

Thoughts?

kris-ms:
@stockholmux I'm cool with everything that you've mentioned here of course. Do you have some ideas on where you'd prefer to have the single source of truth for the guidelines be located? Is there some way of making it easily maintainable so that the designers can make changes when needed? Then we could simply drag that data in for building the page.

The issue this PR is for seemed to be asking for a page to show the guidelines in their current state, so that's my bad for not checking on plans before moving forward.

If you have a spot for the brand assets (and potentially other general assets i.e. author avatars) to live that is not part of this repo I think that the GH actions route could definitely be the go.

PR #159, opened 2024-10-23
adds revises downloads page, adds package page, updates package data

### LINKED ISSUE #131: Moving `artifact_source` to a global setting.
Now that we have the sub-domain setup for `download.valkey.io` and this should not change anymore, we can move the `artifact_source` from specific artifacts' downloads page files(for example [here](https://github.com/valkey-io/valkey-io.github.io/blob/main/content/download/releases/v7-2-6.md?plain=1#L6)) to a global config file `config.toml`

stockholmux:
### Description

This revises. the downloads page extensively.

1) Updates to the package manager information at each version
2) Adds a page with grid of versions and package managers.
3) Moves the `artifact_source` from each release files to a central data file since it never changes

Screenshots:

Package Manager page:
Screenshot 2024-10-23 at 7 52 58 AM

New section on downloads and each release:
Screenshot 2024-10-23 at 7 52 44 AM

### Issues Resolved

#131 

### Check List

### DIFF
--- a/_data/releases.toml
+++ b/_data/releases.toml
@@ -1,0 +1,7 @@
+artifact_source = "https://download.valkey.io/releases/"
+
+[distros]
+
+[distros.notes]
+epel = "Valkey EPEL packages are used in AlmaLinux, Oracle Linux, RHEL (=< 10)"
+
--- a/content/download/packages.md
+++ b/content/download/packages.md
@@ -1,0 +1,11 @@
++++
+template = "packages.html"
+title = "Packages"
++++
+
+You can get Valkey from a variety of package managers in many different operating systems.
+In the chart below, the filled dot (⬤) indicates that the Valkey package for the version in the left most column is available for the package manager indicated in the corresponding column header, the unfilled dot (◯) indicates that the package is not yet available.
+Some package managers may be used across many different distributions (i.e. EPEL with AlmaLinux, RHEL, Oracle Linux, etc. ) or operating systems (i.e. Homebrew or Nix with MacOS and Linux).
+
+This information is maintained on a best-effort basis, if you believe that it is inaccurate, [create an issue](https://github.com/valkey-io/valkey-io.github.io/issues/new/choose).
+Requests for new packages or updates should go to the relevant process for each package manager.
--- a/content/download/releases/v7-2-5.md
+++ b/content/download/releases/v7-2-5.md
@@ -3,7 +3,6 @@ title: "7.2.5"
 date: 2024-04-15
 extra:
     tag: "7.2.5"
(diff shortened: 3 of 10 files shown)


madolson:
Do we really want to be maintaining that dot array? I suppose either:
1. Can we just indicate the minor versions supported. I think that is what is the most useful for what end users support and dramatically reduces the frequency of updates.
2. I wonder if we should just merge it into the Package manager table. Add in the notes, "known support for 8.0".

stockholmux:
I've been investigating the maintenance concerns raised by @madolson.

A lot of this is source from [repology](https://repology.org/project/valkey/packages) we can build up the data using their API and put it into a GitHub Action. Tinkering that I was able to pull the data:

```python
import asyncio
import repology_client
import json


async def main():
    packages = await repology_client.get_packages(project="valkey")
    for package in packages:
        print("---")
        print(package.repo.split("_")[0])
        repo_remainder = package.repo.split("_")[1:]
        print("_".join(repo_remainder))
        print(package.version)
        
asyncio.run(main())
```

This PR relies on getting the availability of Valkey in particular package managers where as repology returns repos with their latest version. It just needs a little extra to transform the data.

avifenesh:
@madolson @zuiderkwast Is it relevant? Should I try to embed it?

I'm not sure if we can make it look nice as it is now, it will be parsing from json to md, then embed the dynamic md as a code block, complicating the status, but feasible.
A lot of the packages on those sites are deprecated, non-official, backports, and we should also add filters to the parsing, and check all kinds of parameters (is it alpine stable? Which version of the distro? Is it a real package, or just named valkey and empty? 8 include the patches? Does it include valkey-cli? Official? And so on...).
The best thing to do, is put a bit of responsibility on users and point clearly that while we're explaining how to do, they should go and check if to do.
I find the idea of trying to maintain problematic, as it may lead to users over-trust the site resources and avoiding checking (which is an easy thing, just go to the link and check your distro so it's not a bad UX).

But if it's relevant, I can take it.

zuiderkwast:
@avifenesh We could use a badge from repology instead? Then we don't need to update it ourselves...

[Packaging status](https://repology.org/project/valkey/versions)

avifenesh:
@zuiderkwast Yes, that looks good, actually. We can leave the instruction of how to, and just add those tags. So it's visible in the same place.
Need to see how to make it more fitted for the site, a grid, or something, but yes, that's good, and not logic and parsing needed.

madolson:
@avifenesh I kind of think we should just put this badge at the bottom of the page with a link.

madolson:
That looks fine to me.

Issue #164, opened 2024-11-22
Updating website with Ubuntu 24.0 Noble support.


roshkhatri:
Recently we have merged this PR https://github.com/valkey-io/valkey/pull/971 for adding support Ubuntu 24.0 Noble release binaries. 

For the next release, we will need to update the website with the same.

stockholmux:
@roshkhatri Next release = patch or minor?

stockholmux:
FWIW, you'll just need to add an object like this to the release file under `artifacts`

```
        -   distro: noble
            arch: 
                -   arm64
                -   x86_64
```

(assuming the binary follows the `https://download.valkey.io/releases/valkey-8.0.1-noble-arm64.tar.gz` pattern)

See:
https://github.com/valkey-io/valkey-io.github.io/blob/main/content/download/releases/v8-0-1.md?plain=1#L22

zuiderkwast:
Now we have "noble" binaries for 8.0, but not for 7.2? I tried and

```
wget https://download.valkey.io/releases/valkey-7.2.8-noble-x86_64.tar.gz
HTTP ERROR response 403
```

What about ubuntu-22 Jammy? No plans for having binary builds for it?

Issue #161, opened 2024-11-19
UpCloud


UP-Will:
**Add Participant Form**

This is an issue created to request the Valkey Project adding a company as a participant in the website. If you want your company to be added please make sure the content is factual and related to the Valkey project.

**Company Information**

- [ ] Provide Company Name: UpCloud
- [ ] Provide Company URL: https://upcloud.com/
- [ ] Provide Description of Company (50 words or less): UpCloud is a leading European cloud provider with a global platform spanning four continents and 13 data centres. Founded in 2012 and headquartered in Helsinki, Finland, our international team ensures uninterrupted services with a 100% SLA, empowering businesses worldwide to thrive with our reliable, cloud-native product portfolio.
- [ ] Provide Link to Valkey related offering: https://upcloud.com/valkey
- [ ] Provide Description related to Valkey offering (50 words or less): Valkey™ on UpCloud simplifies database management with a fully managed service hosted on our GDPR-compliant cloud. Enjoy seamless deployment, expert maintenance, and high performance with minimal downtime. Backed by UpCloud’s global reach, Valkey ensures reliability and scalability, freeing developers to focus on building great applications.
- [ ] Attach the company Logo ![upcloud_logo](https://github.com/user-attachments/assets/262b8f87-c3c5-4187-9887-e55ae65288fa)

**Contact Person**

We want to be able to reach out to you if we have questions while processing this request. Please introduce your contact details (or the details of the responsible person in your company).

- [ ] Ines Pompeu dos Santos
- [ ] INES-SANTOS-UC
- [ ] ines.santos@upcloud.com

stockholmux:
dependent on martinrvisser/valkey-io.github.io#2

stockholmux:
@UP-Will Can you confirm that that #174 looks good? I'll also send [ines.santos@upcloud.com](mailto:ines.santos@upcloud.com) an email.

stockholmux:
closed via #174

PR #166, opened 2024-11-25
Adds a clients overview page

liorsve:
### Description

Adds a client page to docs on the website. The page is added to the top nav docs, called "Client Libraries", and its url is top level `valkey.io/clients `.
This PR is dependent on the [complimentary docs PR](https://github.com/valkey-io/valkey-doc/pull/164) being merged. 

#### How it works - 
- The `init-topics.sh` script was modified to expect the path to the clients docs folder as its second argument. It creates a symlink to it, called `build-clients`. The clients folder in docs holds json files specifying each client's details and features. 
- `content/clients/_index.md` has in its frontmatter the list of paths to the clients appearing in this page. This list is used by the `client-list.html` template to render the specified clients' data. This template also includes the `client-feature-table.html` template that renders the same json files to produce a table comparing which features each client implements. 

### Issues Resolved
[#161](https://github.com/valkey-io/valkey-doc/issues/161)
Dependent on this PR being merged - [#164](https://github.com/valkey-io/valkey-doc/pull/164) 

### Check List


config.toml
```diff
@@ -48,4 +49,16 @@ publish_hold = [
     "/topics/internals/",
     "/topics/protocol/",
     "/topics/rdd/"
+]
+
+recommended_clients_paths = [
```
stockholmux, line 54: Two questions here:
- `config.toml` is a site-level configuration (e.g. included on _every_ page). Is there a better place for this data? I would guess either the macro can load a arbitrary file or include a recommended flag in the JSON file itself. It could also be stored in custom frontmatter (see #167), which might be a flexible alternative to present different recommendations on different pages. 
- How was the 'recommendation' list arrived at?

templates/client_feature_table.html
```diff
@@ -0,0 +1,31 @@
+<hr/>
+<h2>Feature Comparison Table</h2>
+{% if page.slug == "valkey-clients" %}
+    <div style="margin: 20px 0; padding: 10px; border: 2px solid #dcdcdc; background-color: #f9f9f9; border-radius: 5px; overflow-x: auto;">
```
stockholmux, line 4: please use proper CSS classes throughout and define styles with SASS.

Inline styles are difficult to maintain and make consistent style changes across the site.

config.toml
```diff
@@ -48,4 +49,16 @@ publish_hold = [
...
+    "/python/github.com/AWS/valkey-GLIDE.json",
```
stockholmux, line 55: this probably is more relevant for the docs repo but GLIDE is a valkey project, not an AWS project, so the path should be different.

templates/docs-page.html
```diff
@@ -38,11 +38,14 @@ <h1 class="page-title">Documentation: {{ frontmatter_title }}</h1>
 {% else %}
     {{ content_with_fixed_links | markdown | safe }}
 {% endif %}
+{% if page.slug == "valkey-clients" %}
```
stockholmux, line 41: I'm unclear why `page.slug` is being checked here and on `client_feature_table.html`.

On the topic of comparing for the page slug: Is there are more elegant way of doing this? Right now it ties it to a specific page - what if we wanted to include the table on _other_ pages? Ideally this could be achieved without touching the template, only the content.

stockholmux (changes requested):
It needs some adjustments and I have some general approach questions but quite good work on this!

One general comment that I couldn't stick anywhere: Make sure your final PR includes README instructions on how this works.

config.toml
liorsve: Fixed in the docs repo.

build/init-topics.sh
```diff
@@ -42,6 +42,16 @@ done
 
 echo "Topic stub files created at content/topics."
 
+
+#create symlink to clients, expect if it already exists
+topics_path="$1"
+doc_path="${topics_path%%valkey-doc/*}valkey-doc/"
+clients_path="$doc_path/clients"
+#required_client_names=("valkey-glide" "valkey-py" "iovalkey" "valkey-go")
```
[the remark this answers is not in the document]
liorsve: Removed

config.toml
```diff
@@ -21,6 +21,7 @@ highlight_code = true
 [extra]
 command_description_path = "../build-command-docs/"
 command_json_path = "../build-command-json/"
+client_json_path = "../build-clients/"
```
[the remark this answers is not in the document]
liorsve: Added

templates/client_feature_table.html
liorsve: Added, I put it in the _valkey.scss file - hope this is the right place?

templates/docs-page.html
liorsve: No reason, just forgot to remove it from previous versions. Removed it now. 
About comparing for slug, I tried to avoid this and couldn't find another solution, but now that I see #167 , I could maybe create a custom frontmatter with a custom template for this topic that would override the default topic template, which would render all client specific content (the table, but also all of the client list section [as per this comment](https://github.com/valkey-io/valkey-doc/pull/164#discussion_r1858811509)). Does this sound better?

config.toml
liorsve: - Seems like the custom frontmatter is the best option here as it would also help with [avoiding checking for slug](https://github.com/valkey-io/valkey-io.github.io/pull/166#discussion_r1862587794)
- about the list - https://github.com/valkey-io/valkey-doc/pull/164#discussion_r1862496003

liorsve:
- cherry-picked https://github.com/valkey-io/valkey-io.github.io/pull/167
- added a custom frontmatter that overrides the default topics template, and contains the list of recommended clients' paths 
- created a new custom template for the clients page topic, which renders all client specific content from the json client files in valkey-doc/clients (github url, installation and description), and includes the feature comparison table.

config.toml
stockholmux: Alright @liorsve, we're dependent on #164 make sense.

templates/docs-page.html
stockholmux: Yeah, so what we can do is add some flag to `[extra]` section of the front matter (maybe `include_client_table`) and the template checks for that and adds it. Lots of ways to do this, but the general approach is a lot more flexible.

You could even some configuration, e.g. show only python clients or clients that support some feature. Definitely don't need that now, but think about it when implementing something to replace the slug comparison.

zuiderkwast:
I fail to see in the diff what the URL of this page will be.

I suggest we put it under the root, at `https://valkey.io/clients/`. It doesn't really belong under topics.

We already have links to `../clients/` from topics and commands pages, because the corresponding page in the old redis website had this location too. By keeping this, we don't need to update the links in the doc repo.

I don't have a strong opinion about this though. If there is another plan, then we can just update the links in the docs to point to the right path, because I prefer that they make sense for doc writers.

liorsve:
@zuiderkwast  I have no problem moving this, we haven't discussed it and I just put it in topics as a starting point for discussion, so we'll have some kind of visualization. Also it would simplify things if it's not under topics, cause then maybe I won't have to use the topic custom frontmatter PR. Where do you suggest placing this on the website?

templates/client-list.html
```diff
@@ -0,0 +1,67 @@
...
+            <li><a href="#feature-comparison-table">Feature Comparison Table</a></li>
+        </ul>
+        {% for language in languages %}
+            <h2 id="{{ language | slugify }}">{{ language }}</h2>
+            <div class="language-clients">
+                {% for path in page.extra.recommended_clients_paths %}
+                    {% set client = load_data(path = docs::client_json_path(client_path= path), format="json") %}
+                    {% if client.language == language %}
+                    <div class="client-item">
+                        <h3>{{ client.name }}</h3>
```
stockholmux, line 33: needs a `id` with slugified `client.name` as to allow for deep linking to a specific client.

stockholmux (changes requested):
This belongs in the top nav documentation drop down and on the `/docs/` page as well. But the url should be top level (e.g. `valkey.io/clients`

liorsve:
Summarizing the big changes I added - 
1. The page moved from being a topic to the top nav docs, called "Client Libraries". With url `valkey.io/clients`. 
2. Because this is not a topic anymore and it requires creating a symlink to the clients folder in docs, it needs to be included in the shell script. But it's only for creating the symlink and it doesn't create any stub files like the current `init-topics.sh` does for topics, so it's a relatively little addition. So I though it's better to avoid adding a separate `init-clients.sh` file, and instead modify `init-topics.sh` to be `init-topics-and-clients`, and have it expect the path to topics as its first arg, and the path to clients as its second arg. README was updated accordingly.
3. I moved here the markdown content for this page (i.e. explanations about the different client features), which previously was loaded from topics in the docs repo.
4. Updated clients paths according to the change in structure of the clients folder in the docs repo PR - [#164](https://github.com/valkey-io/valkey-doc/pull/164#issuecomment-2580159751).

templates/client-list.html
liorsve: Added

pnbrown:
I'm not able to build this branch.

```
$user valkey-io.github.io % zola serve
Building site...
Checking all internal links with anchors.
> Successfully checked 0 internal link(s) with anchors.
-> Creating 37 pages (0 orphan) and 6 sections
Error: Failed to serve the site
Error: Failed to render section '/Users/niigel/Repos/liorsve/valkey-io.github.io/content/clients/_index.md'
Error: Reason: Failed to render 'client-list.html'
Error: Reason: Function call 'load_data' failed
Error: Reason: `load_data`: /Users/niigel/Repos/liorsve/valkey-io.github.io/../build-clients//python/valkey-GLIDE.json doesn't exist
```

pnbrown:
I see the additional instructions now. Disregard. Will try rebuilding

README.md
```diff
@@ -28,16 +28,16 @@ Changes to external content (command reference, documentation topics) require a
 
 ## Building additional content
 
-**By default, the site will build without documentation topics nor command reference.**
+**By default, the site will build without documentation topics, command reference or the clients page.**
 This content is stored within the `valkey-io/valkey-doc` and `valkey-io/valkey` repos respectively.
```
pnbrown, line 32: Respectively doesn't work here with three items. Also the comma is missing after reference. Suggest:
```
**By default, the site will build without documentation topics, command reference, or the clients page.**
This content for documentation topics and and clients pages are stored within the `valkey-io/valkey-doc` repo. The content for the command reference page is in the `valkey-io/valkey` repo.
```

build/init-topics-and-clients.sh
pnbrown, line 1: There should also be output to the console saying that clients has been built in addition to the line that says topics has been built.

README.md
liorsve: fixed

build/init-topics-and-clients.sh
liorsve: I added an output of "Symlink to clients has been created at ./build-clients" since nothing is actually building here

liorsve:
> The page renders as if the right sidebar is there causing the text to wrap unnecessarily early and the table at the bottom of the page to be squashed. Is there anything that can be done about that?

@pnbrown I updated it to use the fullwidth template, this is what it looks like now. I think the same should work for blogs with a few tweaks.
Screenshot 2025-02-09 at 12 47 58

liorsve:
REQUIRED - merge this PR first that adds the go/valkey-GLIDE file and updates verisons for both glide and valkey-py -

https://github.com/valkey-io/valkey-doc/pull/232

content/clients/_index.md
```diff
@@ -0,0 +1,63 @@
...
+
+This page includes only clients which are regularly tested and recommended. However, it's important to note that all Redis and Valkey clients are compatible with Valkey 7.2 and above. 
+
+Mandatory Features Overview
+----
+1. **Cluster Support** - The ability to operate in a clustered environment, where the data is distributed across multiple shards. Cluster support is essential for applications that require high scalability.
+
+2. **TLS/SSL Support** - The capability to establish secure connections using TLS/SSL, which encrypts the data transmitted between the client and the server. This is a critical feature for applications that require data privacy and protection against eavesdropping.
+
+Advanced Features Overview
```
madolson, line 44: Some of this seems duplicated from the valkey-doc client readme. Maybe we should have an advanced client feature topic and then import it here, so we aren't duplicating it? We can fast follow up with that if you agree, it's not a blocker to merge.

templates/macros/docs.html
```diff
@@ -49,3 +49,17 @@
...
+{%- macro client_json_path(client_path) -%}
+{{config.extra.client_json_path }}{{ client_path }}
+{%- endmacro client_json_path -%}
+
+
+{%- macro boolean_to_checkmark(value) -%}
+    {% if value == true %}
+        ✓
+    {% else %}
+        ✗
```
madolson, line 63: We need some color in this table, otherwise it's very hard to read :) Maybe green if it's implemented and red if it's not?

templates/client-list.html
```diff
@@ -0,0 +1,74 @@
...
+                                                {{ installation_type.type }}:
+                                                <pre>{{ installation_type.command }}</pre>
+                                            </li>
+                                        {% endfor %}
+                                    </ul>
+                                {% else %}
+                                    <code>{{ client.installation }}</code>
+                                {% endif %}
+                            </li>
+                            <li><strong>Version:</strong> <code>{{ client.version }}</code></li>
```
madolson, line 49: Is there a reason the version is code but none of the other fields are? (Installation instructions make some sense, but the other fields could be as well?)

templates/client-list.html
```diff
@@ -0,0 +1,74 @@
...
+                            <li><strong>Version Released:</strong> {{ client.version_released }}</li>
+                            <li><strong>Description:</strong> {{ client.description }}</li>
+                            <li><strong>License:</strong> {{ client.license }}</li>
+                            {% if client.specific_properties %}
```
madolson, line 53: I feel like I missed this from the main PR, but it seems weird looking at this from the main page as only the python libraries have it.

templates/client-list.html
```diff
@@ -0,0 +1,74 @@
+{% extends "fullwidth.html" %}
+{% import "macros/docs.html" as docs %}
+
+{% block main_content %}
```
madolson, line 4: Organizationally, it's weird that you list the advanced features, then list the clients (which doesn't include advanced features), then include at the bottom a table of the advanced features. I think I would either include the advanced features in the client blocks or swap the ordering of the advanced feature description and the client list.

content/clients/_index.md
```diff
@@ -0,0 +1,63 @@
...
+
+This page includes only clients which are regularly tested and recommended. However, it's important to note that all Redis and Valkey clients are compatible with Valkey 7.2 and above. 
+
+Mandatory Features Overview
+----
+1. **Cluster Support** - The ability to operate in a clustered environment, where the data is distributed across multiple shards. Cluster support is essential for applications that require high scalability.
+
+2. **TLS/SSL Support** - The capability to establish secure connections using TLS/SSL, which encrypts the data transmitted between the client and the server. This is a critical feature for applications that require data privacy and protection against eavesdropping.
```
madolson, line 42: Did we decide this someplace? I remember this was in the original description of the PR, but I don't think we finalized on the set of mandatory features. I suppose I would drop this for now and keep the criteria vague until we get wider alignment.

madolson: I think we should also include a link for how you can add your client to this list.

madolson (commented):
looks pretty good

content/clients/_index.md
liorsve: Do you mean an actual topic in docs/topics? Or just dividing the docs/clients README into a separate client feature markdown in the same folder, that we will refer to from the docs README as well?
I think the second option is better, we already create a symlink to the clients docs repo from here so it would be easy to import, and there is no need to create a topic stub file that won't be used afterwards, and also having it as a topic in docs would just create another duplication there. 
I can open a quick PR in the docs repo if you agree, or we can add this post-merge, whatever you prefer

templates/client-list.html
liorsve: Agree, it was added a while ago with the thought that other languages would have specific language features as well but that isn't the case. Removing it

templates/client-list.html
liorsve: Having advanced features per client was the original way it was, but it was too messy so the table was decided as the better option. So i'm swapping the order

content/clients/_index.md
liorsve: > Did we decide this someplace? I remember this was in the original description of the PR, but I don't think we finalized on the set of mandatory features. I suppose I would drop this for now and keep the criteria vague until we get wider alignment.

Will delete this, not sure if this was decided.

> I think we should also include a link for how you can add your client to this list.

Will add, you mean just to refer to docs/clients instructions, right?

content/clients/_index.md
```diff
@@ -0,0 +1,63 @@
...
+    "client_side_caching",
+    "client_capa_redirect",
+    "persistent_connection_pool"
+    ]
+
++++
+
+Selecting the right client is a complex task, given that there are over 200 clients compatible with Valkey across different programming languages. This document offers an overview of recommended Valkey clients for various programming languages. To be included in this list, a client must support a mandatory set of features, such as TLS support and cluster mode. Additionally, a table of advanced features supported by the respective clients is provided, highlighting the unique advantages of one client over another.
+
+This page includes only clients which are regularly tested and recommended. However, it's important to note that all Redis and Valkey clients are compatible with Valkey 7.2 and above.
```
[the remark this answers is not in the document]
liorsve: Changed this to - 
"However, it's important to note that other clients that support Redis OSS version 7.2 are compatible with Valkey 7.2 and above."
Is this ok?

templates/client-list.html
liorsve: Removed code

madolson (approved):
Looks good to me, I noticed that the instructions for valkey-glide go didn't look quite

templates/client-list.html
```diff
@@ -0,0 +1,69 @@
...
+        <ul>              
+            {% set languages = [] %}
+            {% for path in section.extra.recommended_clients_paths %}
+                {% set language = path | split(pat="/") | nth(n=1) %}
+                {% if language not in languages %}
+                    {% set_global languages = languages | concat(with=language) %}
+                {% endif %}
+            {% endfor %}
+            {% for language in languages %}
+            <li><a href="#{{ language | slugify }}">{{ language | title }}</a></li>
```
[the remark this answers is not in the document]
liorsve: - valkey glide- go installation instructions fix PR for docs repo - https://github.com/valkey-io/valkey-doc/pull/241
- Fixed PHP capitalization in the new commit.
-  I also changed the sass style for code width, in the docs PR I added a fix for the glide-java maven and gradle installation which caused it to be wider. After this fix, Gradle installation for Glide now looks like this with a scroll bar - 
Screenshot 2025-03-02 at 10 46 38
But the line too long to fit in the box anyway so I don't see a good way to avoid it.

content/clients/_index.md
```diff
@@ -0,0 +1,59 @@
...
+
+7. **Client Side Caching** - Valkey client-side caching is a feature that allows clients to cache the results of Valkey queries on the client-side, reducing the need for frequent communication with the Valkey server. This can significantly improve application performance by lowering latency, reducing the network usage and cost and reducing the load on the Valkey server. 
+
+8. **`CLIENT CAPA redirect` Support** - The `CLIENT CAPA redirect` feature was introduced in Valkey 8 to facilitate seamless upgrades without causing errors in standalone mode. When enabled, this feature allows the replica to redirect data access commands (both read and write operations) to the primary instance. This ensures uninterrupted service during the upgrade process. For more detailed information about this feature, please refer to [this link.](https://github.com/valkey-io/valkey/pull/325)
```
zuiderkwast, line 57: There are multiple links to PRs.

This is a mistake. We should link to documentation, not to PRs.

madolson: @liorsve Can you take a look?

I was okay referencing PRs, but I'm fine point to actual documentation too.

zuiderkwast: Yeah, it's not a disaster. It's just better to link to the docs. The docs are user facing. The PRs are developer facing.

liorsve: @madolson @zuiderkwast 
I'll open a new PR - we can refer to [docs](https://valkey.io/commands/client-capa/) to explain about client capa from the server side, but to replace the link explaining about AZ awareness the only relevant docs I found is the [blog post about it](https://valkey.io/blog/az-affinity-strategy/) that's mainly focused on Glide, is that ok with you? 
Also, in that PR I can include the follow up to the multiplication you pointed out [here](https://github.com/valkey-io/valkey-io.github.io/pull/166#discussion_r1974516031), LMK if my suggestion there in the comment seems acceptable.

zuiderkwast: Yes, the docs are not very good for this feature. We can keep the link to the PR until we have improved the documentation.

Here is what I found and some ideas:

* We have some documentation of the INFO field `availability_zone` in https://valkey.io/commands/info/ and in the HELLO response in https://valkey.io/commands/hello/
* We don't have the configs in the documentation. This is something I want to have and there is an issue for it: #valkey-io/valkey-doc#154
* We have been talking loosely about creating a document about *client best practices* where we could describe how clients should behave in various situations and maybe also how they should implement features like this one.
* We should probably document reading from replicas properly in https://valkey.io/topics/replication/ or in some other page. The replication page mentions it but it doesn't explain much about it: *"Replication can be used both for scalability, to have multiple replicas for read-only queries (for example, slow O(N) operations can be offloaded to replicas), or simply for improving data safety and high availability."*

PR #169, opened 2024-11-27
Extra valkey in title

stockholmux:
### Description

In valkey-io/valkey-doc#164  @zuiderkwast pointed out that 'Valkey ' was repeated needlessly in the doc titles, leading to silly titles like "Valkey Documentation · Valkey Functions".

This issue resolves the concern by removing "Valkey " from the titles headings on the alphabetical list. 

It smartly recapitalizes the first letter of the title so, for example, a title of "Valkey functions" renders as "Functions" in the alpha list, "Documentation: Functions" in the subhead, and "Valkey Documentation · Functions" in the HTML head `title` tag.

### Issues Resolved

n/a

### Check List

### DIFF
--- a/templates/docs-page.html
+++ b/templates/docs-page.html
@@ -12,7 +12,7 @@
     {% set has_frontmatter = true %}
     {% set page_contents = docs::extract_markdown(content= docs_file_contents) %}
     {% if frontmatter_data.title %}
-        {% set frontmatter_title = frontmatter_data.title %}
+        {% set frontmatter_title = docs::remove_valkey_title(title= frontmatter_data.title) %}
     {% endif %} 
 {% else %}
     {% set page_contents = docs_file_contents %}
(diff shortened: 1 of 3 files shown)


zuiderkwast:
Can't we remove it from each file in the doc repo instead?

zuiderkwast:
If we remove it programatically like this, maybe it makes sense now, but if someone writes a title like "Valkey as a cache" the grammar will break as "Valkey Documentation: As a cache".

stockholmux:
@zuiderkwast If you want to remove it from the source documents, that's fine. My understanding was that you needed for your purpose somehow.

WRT the grammar breaking: yeah, that's a risk. There is a way to catch that on this side via frontmatter override, but it is more optimal to just not have double valkey.

zuiderkwast:
https://github.com/valkey-io/valkey-doc/pull/193

stockholmux:
Issue resolved by https://github.com/valkey-io/valkey-doc/pull/193

Issue #116, opened 2024-08-31
Command syntax: Capitalize tokens


zuiderkwast:
For clarity, I suggest that fixed tokens are capitalized in command syntax. This makes  all fixed content of a command syntax render as capitalized (command name and tokens) and variable content as lowercase (key names, values, etc.).

For example, the *type* tokens of CLIENT LIST.

Current rendering:

    CLIENT LIST [ TYPE normal | master | replica | pubsub ] [ ID client-id ] [ [ ID client-id ] ... ] 

Expected rendering of the `TYPE` argument (with normal, master, replica, pubsub capitalized):

    [ TYPE <NORMAL | MASTER | REPLICA | PUBSUB> ]

Bonus: In this example, it would help to display the `<` and `>` around the arguments after the `TYPE` token, to disambiguate the syntax. Otherwise, `[ TYPE NORMAL | MASTER | REPLICA | PUBSUB ]` can be interpreted as `TYPE NORMAL` or `MASTER` or `REPLICA` or `PUBSUB`, i.e. as if the word "TYPE" is only included in `TYPE NORMAL` but not in the others.

(This is how it's rendered on Redis website and our man pages btw.)

### Details

The TYPE argument in CLIENT LIST has the JSON definition below.

Not that it is "optional" (`[...]` around the whole arguemnt) and has type "oneof" (`<...>` around the things that follow after `TYPE`).

```JSON
            {
                "token": "TYPE",
                "name": "client-type",
                "type": "oneof",
                "optional": true,
                "since": "5.0.0",
                "arguments": [
                    {
                        "name": "normal",
                        "type": "pure-token",
                        "token": "normal"
                    },
                    {
                        "name": "master",
                        "type": "pure-token",
                        "token": "master"
                    },
                    {
                        "name": "replica",
                        "type": "pure-token",
                        "token": "replica"
                    },
                    {
                        "name": "pubsub",
                        "type": "pure-token",
                        "token": "pubsub"
                    }
                ]
            },
```

PR #184, opened 2024-12-22
Use all caps pure-tokens in command "Usage:"

gmbnomis:
In the command page's "Usage:", the lower case `name` was used in case of a `pure-token`. However, the all caps `token` should be used instead. For MIGRATE, this makes an even bigger difference (see `empty-string`):

Before:

```
Usage:
    MIGRATE host port 〈 key | empty-string 〉 destination-db timeout [ copy ] [ replace ] [ AUTH auth | AUTH2 username password ] [ KEYS keys ] [ [ KEYS keys ] ... ]
```

Now:

```
Usage:
    MIGRATE host port 〈 key | "" 〉 destination-db timeout [ COPY ] [ REPLACE ] [ AUTH auth | AUTH2 username password ] [ KEYS keys ] [ [ KEYS keys ] ... ]
```

This addresses a part of #116

### DIFF
--- a/templates/macros/command.html
+++ b/templates/macros/command.html
@@ -11,11 +11,11 @@
             {%- set end = "" -%}
         {%- endif -%}
         {%- set token = "" -%}
-        {%- if arguments.type and arguments.type !=  "pure-token" and arguments.token -%}
+        {%- if arguments.type and arguments.token -%}
             {%- set token = arguments.token -%}
         {%- endif -%}
         {%- set name = "" -%}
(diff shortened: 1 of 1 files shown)


stockholmux (approved):
LGTM

stockholmux:
This looks good. 

I get a little itchy when there are changes to the command reference parser due to the possibility of unintended consequences, but I ran a diff vs the command rendering on current iteration of the website and indeed this appears to only affect spacing (fine/irrelevant) and the intended capitalization.

stockholmux:
@madolson would you mind giving this a review as well?

PR #191, opened 2025-01-12
Command pages: Fix the `fix_links` macro

gmbnomis:
The `fix_links` macro handles markdown strings which do not need HTML escaping. Declaring its template expansion as "safe" prevents double quoting in input like "`<code>`".

This fixes problems when rendering the reply documentation of `CLUSTER INFO`, `GEORADIUS`, and `INFO`.


gmbnomis:
fyi, I diffed the `zola build` output before and after this change and the only relevant changes were the fixes to the command pages mentioned above.

zuiderkwast (approved):
I don't really understand what zola does here, but after some searching i found the templating language and the `safe` filter here:

https://keats.github.io/tera/docs/#safe

In the example for CLUSTER INFO...

``[Bulk string reply](../topics/protocol.md#bulk-strings): A map between named fields and values in the form of `<field>:<value>` lines separated by newlines composed by the two bytes `CRLF`.``


...the HTML tags `<field>:<value>` are not HTML tags but some syntax-like code within backticks, which is get HTML-escaped later when the markdown is rendered as HTML. Is the `{{ ... }}` templating construct actually designed for HTML content, but Zola uses it for markdown text?

PR #195, opened 2025-01-13
Blog post about new hash table

zuiderkwast:
The new new hash table is one of the highlights of the upcoming 8.1 release.

* [x] Improve structure and content of the text
* [x] Replace ascii art with other art?
* [x] Decide which benchmarks we want
* [x] Add benchmark results


content/blog/2025-03-20-new-hash-table.md
```diff
@@ -0,0 +1,202 @@
...
+
+Memory usage for keys of length N and value of length M bytes. TBD.
+
+| Version    | Memory usage per key   |
+|------------|------------------------|
+| Valkey 7.2 | ? bytes                |
+| Valkey 8.0 | ? bytes                |
+| Valkey 8.1 | ? bytes                |
+
+The benchmarks below were run using a key size of N and a value size of M bytes, without pipelining.
```
SoftlyRaining, line 26: Let's add something for set/zset/hash and see if we get even more performance and memory savings since those datatypes are hashtables inside of a hashtable. :)

zuiderkwast: Yes... Feel free to replace these tables with some completely different tests.

There's a fixed overhead for the key and then per field-value. Still I'd like to see a table of memory savings per element/field/etc. for these types.

I want to do hash value embedding (to save the value pointer and an extra allocation) and Ran noticed that our embedded sds (key and field) are sds8 even when they should be sds5, so we could save a two more bytes for those. That's because they're copied from an EMBSTR robj value and those are always sds8. I have some idea to fix that too though.

content/blog/2025-03-20-new-hash-table.md
```diff
@@ -0,0 +1,202 @@
...
+first look for it (due to a hash collission), then it should ideally be found
+within the same cache line. If it is, then it can be found very fast once this
+cache line has been loaded into the CPU cache.
+
+Required features
+-----------------
+
+Why not use an open-source state-of-the-art hash table implementation such as
+Swiss tables? The answer is that we require some specific features, apart from
+the basic operations like add, lookup, replace, delete:
```
SoftlyRaining, line 63: Another reason: Swiss table is very fast, but it stores the elements directly in a contiguous array, which requires that the elements all be the same size. Because our elements vary in size, we had to choose a different design - we chose cache-line sized buckets with element pointers. (This idea was mentioned at the end of the swiss table talk - up to you if you want to make that reference though.)

zuiderkwast: No, you can store pointers in a Swiss table, just like how we store pointers in our bucket layout. The pointers are the fixed-size elements, no?

I don't think it allows a custom key-value entry design like we do though. It can be either a set or a map (key and value) IIUC.

zuiderkwast: I think we could have picked an off-the-shelf implementation even if we couldn't embed key and value the way we do, as long as would be better than dict. It's good to use a battle-tested ready-to-use one too. It's easier to get it right, and less work... I think scan and incremental rehashing were clearly blockers though.

content/blog/2025-03-20-new-hash-table.md
```diff
@@ -0,0 +1,202 @@
++++
+title= "A new hash table"
+date= 2025-03-20 00:00:00
+description= "Designing a state-of-the art hash table implementation"
+authors= [ "zuiderkwast", "SoftlyRaining"]
```
madolson, line 5: Yall need author pages, or it won't render correctly.

content/blog/2025-03-20-new-hash-table.md
```diff
@@ -0,0 +1,202 @@
...
+-------
+
+Memory usage for keys of length N and value of length M bytes. TBD.
+
+| Version    | Memory usage per key   |
+|------------|------------------------|
+| Valkey 7.2 | ? bytes                |
+| Valkey 8.0 | ? bytes                |
+| Valkey 8.1 | ? bytes                |
+
```
madolson, line 25: @SoftlyRaining Are you working on generating these numbers?

content/blog/2025-03-20-new-hash-table.md
```diff
@@ -0,0 +1,202 @@
...
++++
+
+Valkey is essentially a giant hash table attached to the network. A hash table
+is the data structure that maps keys to values. When optimizing for latency, CPU
+and memory usage, it's natural to look at the hash table internals. By replacing
+the hash table with a different implementation, we have managed to reduce the
+memory usage by roughly 20 bytes per key-value pair and improve the latency and
+CPU usage by rougly 10% for instances without I/O threading.
+
+Results
```
madolson, line 15: A typical blog would leave the results till the end, and talk about the constraints up front.

content/blog/2025-03-20-new-hash-table.md
```diff
@@ -0,0 +1,202 @@
...
+The benchmark was run on an xxxx using yyyy, without I/O threads.
+
+Background
+----------
+
+The slowest operation when looking up a key-value pair is by far reading from
+the main RAM memory. A key point when optimizing a hash table is therefore to
+make sure we have as few memory accesses as possible. Ideally, the memory
+reading is already in the CPU cache, which is much faster memory that belong to
+the CPU.
```
madolson, line 43: I think this background is OK, but I think it would be beneficial to show the previous implementation, and then explain how we iterated on it to the new one. Then it becomes easier to show the memory accesses that are getting removed. You already show the previous design later on, so we can probably move it earlier.

A good general framing is, show the current state (the dict), highlight the gaps (lots of memory accesses and allocations), introduce a new concept that solves that (cache friendly), then discuss the solution. Most of that content is here, just needs a little bit of moving around.

content/blog/2025-03-20-new-hash-table.md
```diff
@@ -0,0 +1,202 @@
...
+
+When optimizing for memory usage, we also want to minimize the number of
+allocation and pointers between them, because a pointer is 8 bytes in a 64-bit
+system. If we save one pointer per key-value pair, for 100 million keys that's
+almost a gigabyte.
+
+When a computer loads some data from the main memory into the CPU cache, it does
+so in blocks of one cache line. The cache-line size is 64 bytes on almost all
+modern hardware. Recent work on hash tables, such as "Swiss tables", are highly
```
madolson, line 52: Should probably link the swiss table design notes or youtube video here.

content/blog/2025-03-20-new-hash-table.md
```diff
@@ -0,0 +1,202 @@
...
+optimized for cache lines. When looking up a key, if it's not found where you
```
madolson, line 53:
```suggestion
modern hardware. Recent work on hash tables, such as "Swiss tables", are highly
optimized to store and access data within a single cache line. When looking up a key, if it's not found where you
```

content/blog/2025-03-20-new-hash-table.md
```diff
@@ -0,0 +1,202 @@
...
+first look for it (due to a hash collission), then it should ideally be found
```
madolson, line 54:
```suggestion
first look for it (due to a hash collision), then it should ideally be found
```
First introduction of concepts such as hash collisions. I mentioned earlier it would be helpful to give an overview of the current design, probably can walk through these concepts.

content/blog/2025-03-20-new-hash-table.md
```diff
@@ -0,0 +1,202 @@
...
+the table, their key-value entries form a linked list. That's what the "next"
+pointer in the dictEntry is for.
+
+To lookup a key "FOO" and access the value "BAR", Valkey still had to read from
+memory four times. If there is a hash collission, it has to follow two more
+pointers for each hash collission and thus read twice more from memory (the key
+and the next pointer).
+
+In Valkey 8.0, an optimization was made to embed the key ("FOO" in the drawing
+above) in the dictEntry, eliminating one pointer and one memory access.
```
madolson, line 124: We have a full blog post dedicated to embedding the key, so let's just start with the state of the world in 7.2.

content/blog/2025-03-20-new-hash-table.md
```diff
@@ -0,0 +1,202 @@
++++
+title= "A new hash table"
+date= 2025-03-20 00:00:00
+description= "Designing a state-of-the art hash table implementation"
+authors= [ "zuiderkwast", "SoftlyRaining"]
++++
+
+Valkey is essentially a giant hash table attached to the network. A hash table
```
madolson, line 8: Probably want a better hook. We can talk about how many caching workloads are bound on storing data, so being able to to store more data allows you to reduce the size of your clusters.

content/blog/2025-03-20-new-hash-table.md
```diff
@@ -0,0 +1,202 @@
...
+In Valkey 8.0, an optimization was made to embed the key ("FOO" in the drawing
+above) in the dictEntry, eliminating one pointer and one memory access.
+
+Design
+------
+
+In the new hash table designed for Valkey 8.1, the table consists of buckets of
+64 bytes, one cache line. Each bucket can store up to seven elements. Keys that
+map to the same bucket are all stored in the same bucket. The bucket also has a
+metadata section which contains a one byte secondary hash for each key. This is
```
madolson, line 132: Need to explain a bit about what the secondary hash is doing here, it's not really explained what it is or how it solve hash collisions.

I'm not sure we did any testing, but this is also really good for caching workloads, since we save a memory access most of the time when there is a cache miss. A common cache hit rate is like 80%, so that 20% gets a nice speed boost as well.

content/blog/2025-03-20-new-hash-table.md
```diff
@@ -0,0 +1,202 @@
...
+
+Assuming the hashtable and the table are already in the CPU cache, looking up
+key-value entry now requires only two memory lookups: The bucket and the
+serverObject. If there is a hash collission, the object we're looking for is
+most likely in the same bucket, so no extra memory access is required.
+
+If a bucket becomes full, the last element slots in the bucket is replaced by a
+pointer to a child bucket. A child bucket has the same layout as a regular
+bucket, but it's a separate allocation. Child buckets form a chain. There is
+some probability of this happening, but long chains are very rare. Most of the
```
madolson, line 165:
```suggestion
bucket, but it's a separate allocation. The length of these bucket chains are not bounded, but long chains are very rare as long as keys are well distributed by the hashing function. Most of the
```

The, "there is some probability" just seems weird to me.

content/blog/2025-03-20-new-hash-table.md
```diff
@@ -0,0 +1,202 @@
...
+
+Hashes, sets and sorted sets
+----------------------------
+
+The nested data types Hashes, Sets and Sorted sets also make use of the new hash
+table. The memory usage is reduced by roughly 10-20 bytes per entry. Memory and
+latency/throughput results are WIP.
+
+Iterator prefetching
+--------------------
```
madolson, line 194: This is technically a separate point. It addresses some of the content of the blog, but ultimately feels a bit random at the end. Maybe we drop it and write another followup blog on memory prefetching?

madolson (commented):
Overall I think the content is really good. Is @SoftlyRaining planning on adding more information about benchmark results?

content/blog/2025-03-20-new-hash-table.md
zuiderkwast: Just move the Results section to the end?

I'll keep brief text in the abstract above *"we have managed to reduce the memory usage by roughly 20 bytes per key-value pair and improve the latency and CPU usage by rougly 10% for instances without I/O threading"*?

content/blog/2025-03-20-new-hash-table.md
zuiderkwast: Yeah, it's hard to know how much of computer science lecture we have to give. We'll have to assume some basic knowledge of what a hash table is, right?

content/blog/2025-03-20-new-hash-table.md
zuiderkwast: We could include it in the 8.1 overview blog post. I'm not sure anyone wants to write a prefetching blog post specifically.

content/blog/2025-03-20-new-hash-table.md
zuiderkwast: Thanks. I'll take this with some line breaks. (I don't like long lines. :laughing:)

```suggestion
bucket, but it's a separate allocation. The length of these bucket chains are
not bounded, but long chains are very rare as long as keys are well distributed
by the hashing function. Most of the
```

zuiderkwast:
@rjd15372 Do you want to take a look?

content/blog/2025-03-28-new-hash-table/index.md
```diff
@@ -0,0 +1,229 @@
...
+
+The dict has two tables, called "table 0" and "table 1". Usually only one
+exists, but both are used when incremental rehashing is in progress.
+
+It's a chained hash table, so if multiple keys are hashed to the same slot in
+the table, their key-value entries form a linked list. That's what the "next"
+pointer in the `dictEntry` is for.
+
+To lookup a key "FOO" and access the value "BAR", Valkey still has to read from
+memory four times. If there is a hash collission, it has to follow two more
```
rjd15372, line 69:
```suggestion
memory four times. If there is a hash collision, it has to follow two more
```

content/blog/2025-03-28-new-hash-table/index.md
```diff
@@ -0,0 +1,229 @@
...
+pointers for each hash collission and thus read twice more from memory (the key
```
rjd15372, line 70:
```suggestion
pointers for each hash collision and thus read twice more from memory (the key
```

content/blog/2025-03-28-new-hash-table/index.md
```diff
@@ -0,0 +1,229 @@
...
+Results
+-------
+
+By replacing the hash table with a different implementation, we've managed to
+reduce the memory usage by roughly 20 bytes per key-value pair.
+
+The graph below shows the memory overhead for different value sizes. The
+overhead is the memory usage excluding the key and the value itself. Lower is
+better. The zigzag pattern is because of aliasing between the datapoint spacing
+and the memory allocator's discrete allocation sizes.
```
rjd15372, line 207: "The zigzag pattern is because of aliasing between the datapoint spacing
and the memory allocator's discrete allocation sizes."

I couldn't understand what you mean with this. Can you describe it in another way to make it more clear?

zuiderkwast: I took this from Rain's issue with the diagrams. :)

It's wasted memory when jemalloc allocates more memory than we need. The sizes are as in this table: http://jemalloc.net/jemalloc.3.html#size_classes

Explaining this is probably not worth it. I'll cut it down to this, ok?

> The zigzag pattern is because of the memory allocator's discrete
allocation sizes.

content/blog/2025-03-28-new-hash-table/index.md
```diff
@@ -0,0 +1,166 @@
...
++++
+
+Many workloads are bound on storing data. Being able to store more data using
+less memory allows you to reduce the size of your clusters.
+
+In Valkey, keys and values are stored in what's called a hash table. A hash
+table works by chopping a key into a number of seemingly random bits. These bits
+are shaped into a memory address, pointing to where the value is supposed to be
+stored. It's a very fast way of jumping directly to the right place in memory
+without scanning trough all the keys.
```
madolson, line 15:
```suggestion
without scanning through all the keys.
```

content/blog/2025-03-28-new-hash-table/index.md
```diff
@@ -0,0 +1,166 @@
...
+![hashtable structure](hashtable-structure.png)
+
+Assuming the `hashtable` structure is already in the CPU cache, looking up
+key-value entry now requires only two memory lookups: The bucket and the
+`serverObject`. If there is a hash collision, the object we're looking for is
+most likely in the same bucket, so no extra memory access is required.
+
+If a bucket becomes full, the last element slot in the bucket is replaced by a
+pointer to a child bucket. A child bucket has the same layout as a regular
+bucket, but it's a separate allocation. The length of these bucket chains are
```
madolson, line 105:
```suggestion
bucket, but it's a separate allocation. The lengths of these bucket chains are
```

content/blog/2025-03-28-new-hash-table/index.md
```diff
@@ -0,0 +1,166 @@
...
+![memory usage by version](memory-usage.png)
+
+For keys with an [expire time](/commands/expire/) (time-to-live, TTL) the memory
+usage is down even more, roughly 30 bytes per key-value pair.
+
+![memory usage with expire](memory-usage-with-expire.png)
+
+In some workloads, such as when storing very small objects and when pipelining
+is used extensively, the latency and CPU usage are also improved. In most cases
+though this is negligble in practice. The key takeaway appears to be reduced
```
madolson, line 155:
```suggestion
though this is negligible in practice. The key takeaway appears to be reduced
```

madolson (commented):
Pass for grammar, will auto-apply these changes

madolson (approved):
I like it.

PR #201, opened 2025-02-02
Implement a properly sorted downloads page

madolson:
### Description

Added a quick and dirty sorting mechanism for sorting the downloads page. Includes the latest version from each major, with a new sorting based of a weight, which is a numeric representation of the semantic version.

Notes that weights must be a positive number, so you have to do a reverse sort and can't use a negative number for the version.

image
 
### Check List

### DIFF
--- a/content/download/releases/_index.md
+++ b/content/download/releases/_index.md
@@ -1,5 +1,5 @@
 +++
 page_template= "release-page.html"
 template= "release-section.html"
-sort_by= "weight"
+sort_by= "title"
 +++
--- a/content/download/releases/_index.md
+++ b/content/download/releases/_index.md
@@ -1,5 +1,5 @@
 +++
 page_template= "release-page.html"
 template= "release-section.html"
-sort_by= "date"
+sort_by= "weight"
 +++
--- a/templates/download.html
+++ b/templates/download.html
@@ -7,15 +7,26 @@
 
 
 {% set release_section = get_section(path="download/releases/_index.md") %}
-{% set most_recent_release_page = release_section.pages | slice(end= 1)  %}
+{% set most_recent_release_page = release_section.pages | reverse | slice(end= 1) %}
 {% set release_date = most_recent_release_page[0].date %}
 {% set release = most_recent_release_page[0].extra %}
 
-You can get the latest release of Valkey ({{ release.tag }}) from this page.<br />
-The <a href="./releases/">releases</a> page contains links to download all current and previous releases (including any security fixes for previous released versions).<br />
-<br />
-
-<hr />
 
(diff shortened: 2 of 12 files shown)


zuiderkwast (approved):
It can't sort on strings? The title field contains the version as a string.

content/download/releases/_index.md
```diff
@@ -1,5 +1,5 @@
 +++
 page_template= "release-page.html"
 template= "release-section.html"
-sort_by= "date"
+sort_by= "weight"
```
zuiderkwast, line 4: Yeah, I tested locally. Sort lexicographically by title works. With this, we don't need to add the weight field.

```suggestion
sort_by= "title"
```

templates/download.html
```diff
@@ -7,15 +7,33 @@ <h1 class="page-title">Download Latest</h1>
...
+You can get the latest releases of Valkey from this page.<br />
 The <a href="./releases/">releases</a> page contains links to download all current and previous releases (including any security fixes for previous released versions).<br />
 <br />
 
 <hr />
 
+<h2> Version {{major}}.x.x </h2>
 {% include "includes/release.html" %}
+
+<h1> Latest supported version of past releases</h1>
```
zuiderkwast, line 26: It's not good to have more than one H1 on the same page. SEO stuff and bad typography.

templates/download.html
```diff
@@ -7,15 +7,33 @@ <h1 class="page-title">Download Latest</h1>
...
+{% set_global major = split_ver | nth(n= 0) %}
+
+You can get the latest releases of Valkey from this page.<br />
 The <a href="./releases/">releases</a> page contains links to download all current and previous releases (including any security fixes for previous released versions).<br />
 <br />
 
 <hr />
 
+<h2> Version {{major}}.x.x </h2>
 {% include "includes/release.html" %}
```
zuiderkwast, line 24: The file "includes/release.html" prints a number of `<h2>` headings. When we print multiple versions on this page, I think it's better to change all `<h2>` in the include template to `<h3>`. Then we can have h2 for the version title and h3 for the stuff under it (docker hub, binary artifacts).

madolson: Sure, I'm okay with that.

content/download/releases/_index.md
madolson: So, lexographic won't work because 8.0.2 will come after 8.0.10, but apparently sorting by title isn't actually lexographic. "Zola uses a natural sorting algorithm which means that titles like "Chapter 2" will be sorted correctly before "Chapter 10"." cool

zuiderkwast (approved):

PR #208, opened 2025-02-17
Automatically deploy website on updates to main

madolson:
### Description

Automatically deploy the website when we update main, instead of having to merge it to Zola.
 
### Issues Resolved

No issue, but discussed it offline.

### Check List

### DIFF
--- a/.github/workflows/zola-deploy.yml
+++ b/.github/workflows/zola-deploy.yml
@@ -1,8 +1,8 @@
-name: Zola on GitHub Pages
+name: Deploy website
 
 on:
   push:
-    branches: ["prod_zola"]
+    branches: ["main"]
   
   workflow_dispatch:
 
@@ -18,7 +18,7 @@ jobs:
       - name: Checkout prod
         uses: actions/checkout@v4
         with:
-          ref: prod_zola
+          ref: main
           path: website
 
       - name: Checkout valkey-doc


stockholmux:
I'd advise you keep some sort of staging. I've seen too many accidents that go directly to production (😬), plus there are times where you need to do a last minute change for things like publish dates.


(I'll go back on leave now)

madolson:
>I'd advise you keep some sort of staging. I've seen too many accidents that go directly to production (😬), plus there are times where you need to do a last minute change for things like publish dates.

Hehe, hope you are enjoying your time off! We talked about it a bunch and found it just caused us not to deploy quickly. Although I agree conceptually, I'm not sure if there is another good staging mechanism.

zuiderkwast (approved):
Yeah we discussed it.

We have many errors on the website such as broken links to downloads and such that are not fixed because of the extra steps to deploy.

If zola build fails, it isn't deployed, so there is no harm.

There were some more points but i don't remember now.

madolson:
> Hehe, hope you are enjoying your time off! We talked about it a bunch and found it just caused us not to deploy quickly. Although I agree conceptually, I'm not sure if there is another good staging mechanism.

Just to clarify since someone mentioned concerns about an offline discussion. I just summarized the output which is that we keep not-deploying the website because we have a secondary process to move to production. I can't speak extensively for others, but people don't check the staging system all that closely, and we want to keep adding more validation that is automated and avoiding manual interventions.

We're still a project that needs to move fast, and this is a two way door decision, since it's easy to revert.

PR #209, opened 2025-02-18
Add 8.1.0-rc1 release

ranshid:


content/download/releases/v8-1-0.md
```diff
@@ -0,0 +1,34 @@
+---
+title: "8.1.0-rc1"
```
madolson, line 2: This will make Valkey 8.1 show up as a the "latest" 8.0 release, which is not what we want. We need to make some change so that it is ignored.

madolson: Here is a fix for it: https://github.com/valkey-io/valkey-io.github.io/pull/210

madolson (approved):

PR #233, opened 2025-04-02
feat(web): migrate website to React

phiro56:
This pull request delivers a complete UI/UX redesign of the client-facing website using React. The goal is to modernize the visual identity and improve usability.

### Description
**Migrate website to React and new Design**
- Replace with React application
- Reorganize documentation structure with new MD files
- Move static assets to public directory
- Add necessary configuration for React development (eslint, prettier)

### ⚠️ Notes
- This is a breaking change: the old frontend is fully replaced


microwavables:
> This is a giant change. Can you say more why you moved to React?
> 
> On the surface, I do not believe this good move.

@stockholmux  Kyle, can you please elaborate why you do not believe this is a good move?

zuiderkwast:
Does this mean we'll no longer be using Zola? The change is super huge and all the content has been changed/moved so I assume this is the case.

People have just started to learn how it works and we've been fixing annoyances of the year such as much of how the command pages are rendered. If we move to something else now, we'll have to start fixing things like that from scratch. So I must say I'm really against it.

We've been informed that there's a new web design in the works and that's what we've said yes to. To me, that means changing some CSS. Not the content nor the logic for rendering the HTML.

stockholmux:
@microwavables React is a *web application* framework. Valkey.io is clearly a web site: it is designed to serve people information, not something that is an interactive application. 

Why is this a big deal? Well, it adds very non-trivial complexity to the site and the advantages are suspect. 

The complexity is from a bunch of different angles. I think the biggest is the requirements to do development on the website will go up steeply. [This change requires](https://github.com/phiro56/valkey-io.github.io/blob/feat-web-ui-redesign/package.json) installing package managers (npm), a new runtime (node), a transpiler (typescript), etc. 

Zola was a single binary. 

To give some context: this site was originally written in Jekyll but moved to Zola because multiple people failed to contribute because they found installing Ruby and the related package manager too onerous. IMHO, React is _far_ more difficult than getting Jekyll up and running. 

If Valkey was a community of frontend folks, React would probably be fine. The reality is that Valkey developers and users are far removed from this type of tooling and it's a big lift and ask to install all this stuff to, say, test out how documentation looks or write a blog post.

(Additionally, if you look back in the issues on the repo, React was already discussed and didn't move forward, multiple people saying 'nah'.) 

This doesn't even mention the impact on users: I haven't had the time to evaluate the end product of this PR but typically React sites are quite heavy and negatively impact performance and suffer badly on poor connections.

makubo-aws:
I think we're aligned on keeping the site static and will migrate from React to a static generator.

PR #236, opened 2025-04-16
add AZ_AFFINITY_REPLICAS_AND_PRIMARY read strategy example and diagram

mo-amzn:
### Description
Adding documentation for the new read strategy AzAffinityReplicasAndPrimary
relates to issues https://github.com/valkey-io/valkey-glide/issues/2792 and https://github.com/valkey-io/valkey-glide/issues/3085

 
### Check List


Muhammad-awawdi-amazon:
@stockholmux @madolson 
can I get a review please?

adarovadya:
add yourself to the authors as well

content/blog/2024-12-22-az-affinity-strategy.md
```diff
@@ -28,6 +28,7 @@ GLIDE provides flexible options tailored to your application’s needs:
 * ```PRIMARY```: Always read from the primary to ensure the freshness of data.
 * ```PREFER_REPLICA```: Distribute requests among all replicas in a round-robin manner. If no replica is available, fallback to the primary.
 * ```AZ_AFFINITY```: Prioritize replicas in the same AZ as the client. If no replicas are available in the zone, fallback to other replicas or the primary if needed.
+* ```AZ_AFFINITY_REPLICAS_AND_PRIMARY```: Prioritize replicas in the same AZ as the client. If no replicas are available in the zone, fallback to the primary in the same AZ. If neither are available, fallback to other replicas or the primary in other zones.
 
 In Valkey 8,  ```availability-zone``` configuration was introduced, allowing clients to specify the AZ for each Valkey server. GLIDE leverages this new configuration to empower its users with the ability to use AZ Affinity routing. At the time of writing, GLIDE is the only Valkey client library supporting the AZ Affinity strategy, offering a unique advantage.
```
adarovadya, line 33: change it to ```strategies``` instead strategy

content/blog/2024-12-22-az-affinity-strategy.md
```diff
@@ -53,6 +54,20 @@ In Valkey 8,  ```availability-zone``` configuration was introduced, allowing cli
...
     ![AZ_AFFINITY Read strategy latency example](/assets/media/pictures/AZ_AFFINITY_strategy.png)
+    
+    **Example 2:**
+    Consider a cluster with three nodes, one primary and two replicas. Each node is located in a different availability zone. The client located in az-2 along with the primary node. The replicas are located in az-1 and az-3.
+    
+    **With ```AZ_AFFINITY``` strategy**:
+    In this case, the client attempts to read from a replica in the same AZ. Since none are available in az-2, it falls back to a replica in another AZ, such as az-1 or az-3.  
+    and the average latency is, for example, about 800 microseconds.
+    
+    ![AZ_AFFINITY Read strategy latency example](/assets/media/pictures/AZ_AFFINITY_strategy2.png)
```
adarovadya, line 65: I think it's confusing that in the first example the "super" client is the AZ_AFFINITY user, but later the "super" client becomes the AZ_AFFINITY_REPLICAS_AND_PRIMARY user.
I suggest having **one example** that clearly shows all three cases, and explaining when a user should prefer AZ_AFFINITY over AZ_AFFINITY_REPLICAS_AND_PRIMARY, and vice versa.

Muhammad-awawdi-amazon: I see this a bit differently. The new example visually clarifies the differences between AZ_AFFINITY and AZ_AFFINITY_REPLICAS_AND_PRIMARY, illustrating real-world scenarios where each one fits best. I'd prefer to retain both examples.

content/blog/2024-12-22-az-affinity-strategy.md
```diff
@@ -28,8 +28,9 @@ GLIDE provides flexible options tailored to your application’s needs:
 * ```PRIMARY```: Always read from the primary to ensure the freshness of data.
 * ```PREFER_REPLICA```: Distribute requests among all replicas in a round-robin manner. If no replica is available, fallback to the primary.
 * ```AZ_AFFINITY```: Prioritize replicas in the same AZ as the client. If no replicas are available in the zone, fallback to other replicas or the primary if needed.
+* ```AZ_AFFINITY_REPLICAS_AND_PRIMARY```: Prioritize replicas in the same AZ as the client. If no replicas are available in the zone, fallback to the primary in the same AZ. If neither are available, fallback to other replicas or the primary in other zones.
```
stockholmux, line 31: My general thoughts on updating blogs is... don't. Typically, blogs represent a slice-in-time picture of the technology. However, if you feel like it's important to update, I would add a small note (maybe in italics) at the bottom that the blog post was updated: include what you added and when. Something like "This blog was updated in May 2025 to include `AZ_AFFINITY_REPLICAS_AND_PRIMARY` and related example"

stockholmux (changes requested):
I'm generally fine with the content, but I think you need to add an update notice at the bottom of the blog.

content/blog/2024-12-22-az-affinity-strategy.md
Muhammad-awawdi-amazon: Since we often attach this blog and send to users/customers, we prefer that it stays up-to-date 
I added a notice regarding the blog update

stockholmux (approved):
LGTM

content/blog/2024-12-22-az-affinity-strategy.md
stockholmux: @mo-amzn If you're sending folks here, it's serving as documentation and this should probably migrate to documentation.

adarovadya (approved):

PR #243, opened 2025-04-22
Bug fix: allow local builds without requiring client directory

liorsve:
### Description

Currently building the zola site locally is not possible when the clients directory is not built, as the `zola serve` command fails.
This is fixed by introducing an additional boolean variable `clients_built` in the client's page frontmatter, which is set to false by default, but updates to be true when running the `init_topics_and_clients.sh` script successfully. The page loads only if this var is set to true, otherwise a "Clients page not found" message is displayed, similar to the behavior in the commands and topics documentation.
 
### Check List

### DIFF
--- a/templates/client-list.html
+++ b/templates/client-list.html
@@ -2,68 +2,74 @@
 {% import "macros/docs.html" as docs %}
 
 {% block main_content %}
-    <div class="markdown-content">
-        {% set content_parts = section.content | split(pat="<!-- split -->") %}
-        {{ content_parts[0] | markdown | safe }}
-    </div>
+    {% set client_path = section.extra.recommended_clients_paths | first %}
+    {% set client = load_data(path = docs::client_json_path(client_path=client_path), format="json", required = false) %}
+    {% if not client %}
+        <strong>Clients page not found.</strong> You likely need to build the client page. See "Building additional content" in the README.md file.
+    {% else %}
+        <div class="markdown-content">
+            {% set content_parts = section.content | split(pat="<!-- split -->") %}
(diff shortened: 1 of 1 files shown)


content/clients/_index.md
```diff
@@ -2,6 +2,7 @@
 title = "Client Libraries"
 template = "client-list.html"
 [extra]
+clients_built = true
```
stockholmux, line 5: Shouldn't this be set to `false` initially?

stockholmux (commented):
So, I can't get this to work as written.

If I change the value of `clients_built`  to `false` then it does indeed not need the clients to build successfully (yay!) but if then try to run `./build/init-topics-and-clients.sh ../valkey-doc/topics ../valkey-doc/clients` I get `sed: 1: "content/clients/_index.md": command c expects \ followed by text` and it doesn't flip the front matter boolean.

stockholmux:
> @stockholmux did you run it on macOS? I’ve updated the `sed` command so it works on both Linux and macOS now, in case that was the issue

Yep. I was testing on macOS. Any other flavours of `sed` we need to worry about? 

The other thought that comes to mind now is if we're going to have constant problems with people accidentally committing `clients_built = true`to `content/clients/_index.md`

liorsve:
So, I enclosed all of the content of this page with a check at the beginning - trying to use `load_data` with the first client's path and with `required = false`, and if it doesn't exist displaying the `page not found` message. It works for me locally without the extra changes to the other `load_data` calls.

stockholmux (approved):
LGTM!

PR #255, opened 2025-05-04
added Try Valkey support

zarkash-aws:
### General Description

This PR implements a browser-based environment that allows users to interact with Valkey-CLI commands without needing local installation. This feature was initially introduced in [Issue #1412](https://github.com/valkey-io/valkey/issues/1412), and it aims to provide a lightweight, easy-to-access platform for users to explore Valkey’s functionalities directly in the browser.

### Implementation 

The solution uses a virtualized environment powered by [v86](https://github.com/copy/v86), a tool that emulates an x86-compatible CPU and hardware. The machine code for the environment is translated into WebAssembly modules at runtime, allowing it to run entirely within the browser.

* VM Configuration: When the page is loaded, the VM loads a preconfigured Alpine Linux image that already has the Valkey server and CLI installed. This configuration ensures users can immediately interact with the Valkey CLI upon starting the web-based environment.
* Current Version: The current version of Valkey in the preconfigured image is 8.1.0. However, binaries for earlier versions (7.2.8 and 8.0.1) are also available in the S3 bucket. We can decide whether to set a single version or implement a dropdown menu that allows users to choose the version they want to interact with.

### Serving of Binary Files

The virtual machine used for this setup relies on two main binary file components:

1. Binary files for the VM’s filesystem.
2. A binary file containing a preloaded state, which allows for seamless page loading.

These binaries are served from an S3 bucket (currently set as a test bucket) via CloudFront CDN.

* CORS Policy: To allow the browser to access these binary files, we need to configure the appropriate CORS policy. Specifically, the “Access-Control-Allow-Origin” header must be set correctly. This CORS policy needs to be configured both for the S3 bucket and the CloudFront distribution.
    * We can set this policy only for the Try-Valkey files directory if we prefer not to apply it to the entire bucket.
    * Additionally, we can restrict access to these files by specifying that only the Valkey website domain can use them. More details can be found [here](https://repost.aws/knowledge-center/no-access-control-allow-origin-error).

### New Image Creation Process

Currently, the creation of new images for future Valkey versions will be done manually. However, we plan to automate this process in the future to simplify updates as new versions of Valkey are released.
I have also created a [repository](https://github.com/valkey-io/valkey-try-me) for documentation and for creating new images for Try Valkey. 

### Screenshots
new "Try Valkey" tab
Screenshot 2025-06-03 at 11 48 36
data warning 
Screenshot 2025-06-03 at 11 48 47
after pressing "load emulator"
Screenshot 2025-06-03 at 11 49 00

### DIFF
--- a/templates/valkey-try-me.html
+++ b/templates/valkey-try-me.html
@@ -1,7 +1,5 @@
 {% extends "fullwidth.html" %}
-{%- block head -%}
-    <title>Try Valkey</title>
-
+{% block main_content %}
     <!-- Scripts -->
     <script src="https://download.valkey.io/try-me-valkey/vos/v86/libv86.js"></script>
     <script src="https://download.valkey.io/try-me-valkey/vos/xterm/xterm.min.js"></script>
@@ -11,17 +9,17 @@
     <!-- Styles -->
     <link rel="stylesheet" href="https://download.valkey.io/try-me-valkey/vos/xterm/xterm.css" />
     <link rel="stylesheet" href="https://download.valkey.io/try-me-valkey/vos/valkey-try-me.css" />
-
-{%- endblock -%}
-{% block main_content %}
-    <p>This is an in-browser Valkey server and CLI that runs directly within your browser using a V86 emulator, requiring no external installations. </p>
+    
+    <!-- Body -->
+    <title>Try Valkey</title>
+    <p>This is an in-browser Valkey server and CLI that runs directly within your browser using a <a href="https://github.com/copy/v86">V86</a> emulator, requiring no external installations. </p>
     <p>Try it out below:</p>
     <div id="terminalWrapper" class="container" style="display: none;">
         <div id="terminal-container"></div>
     </div>
     <!-- Warning Section -->
     <div id="warningContainer" style="text-align: center; margin-top: 20px;">
-        <button id="startButton" style="padding: 10px 20px; font-size: 18px; margin-top: 10px;">Load Emulator</button>
+        <button id="startButton" style="padding: 10px 20px; font-size: 18px; margin-top: 10px; cursor: pointer;">Load Emulator</button>
         <p>This emulator will download approximately 50MB of data.</p>
     </div>
     <!-- Loading Section (Hidden at first) -->
@@ -29,8 +27,6 @@
         <p id="progressText">Preparing to load...</p>
         <progress id="progressBar" value="0" max="100"></progress>
     </div>
-    
-
 
     <script>
         "use strict";
--- a/templates/valkey-try-me.html
+++ b/templates/valkey-try-me.html
@@ -3,56 +3,15 @@
     <title>Try Valkey</title>
 
     <!-- Scripts -->
-    <script src="https://dtq9kvai1345o.cloudfront.net/vos/v86/libv86.js"></script>
-    <script src="https://dtq9kvai1345o.cloudfront.net/vos/xterm/xterm.min.js"></script>
-    <script src="https://dtq9kvai1345o.cloudfront.net/vos/pako/pako.min.js"></script>
-    <script src="https://dtq9kvai1345o.cloudfront.net/vos/v86/serial_xterm.js"></script>
+    <script src="http://dtq9kvai1345o.cloudfront.net/vos/v86/libv86.js"></script>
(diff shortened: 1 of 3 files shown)


templates/valkey-try-me.html
```diff
@@ -0,0 +1,277 @@
+{% extends "fullwidth.html" %}
+{%- block head -%}
+    <title>Try Valkey</title>
+
+    <!-- Scripts -->
+    <script src="https://dtq9kvai1345o.cloudfront.net/vos/v86/libv86.js"></script>
+    <script src="https://dtq9kvai1345o.cloudfront.net/vos/xterm/xterm.min.js"></script>
+    <script src="https://dtq9kvai1345o.cloudfront.net/vos/pako/pako.min.js"></script>
+    <script src="https://dtq9kvai1345o.cloudfront.net/vos/v86/serial_xterm.js"></script>
```
madolson, line 9: Are these things we can host?

templates/valkey-try-me.html
```diff
@@ -0,0 +1,277 @@
...
+    
+    <!-- Styles -->
+    <link rel="stylesheet" href="https://dtq9kvai1345o.cloudfront.net/vos/xterm/xterm.css" />
```
madolson, line 12: Can we host this just on the website?

templates/valkey-try-me.html
```diff
@@ -0,0 +1,277 @@
...
+
+    <style>
```
madolson, line 14: Let's split this off in to a dedicated CSS file. I'm told embedded CSS is worse for performance.

madolson:
Is this a draft or ready to review?

templates/valkey-try-me.html
ranshid: yes. The current plan is that we will manually upload the binaries created using [this](https://github.com/valkey-io/valkey-try-me) repository. the automation is something that will reqiure some extra effort we decided to wait with. It is a perquisite to have them loaded to our production s3 bucket before we merge this PR.

templates/valkey-try-me.html
ranshid: I think that would drive some features to be implemented cross repositories and we should avoid that. I agree the fact that the css files are more visualisation related might seem rational to include in the webpage repo, but since they are somewhat correlated with the v86 objects I would suggest to avoid it at this point.

templates/valkey-try-me.html
ranshid: I am not sure I understand the idea - can explain which part should we split?

ranshid:
> Is this a draft or ready to review?

@madolson currently the content is pointing a test bucket with preloaded binaries. I think the PR should be reviewed after @zarkash-aws will change it to point to our production bucket

templates/valkey-try-me.html
roshkhatri: I think we can move these files to the production S3 bucket and add those links here.

zarkash-aws: the files were moved to the production s3 bucket, see latest commit.

roshkhatri (approved):
over all the PR LGTM
@zarkash-aws Can you please render this PR and attach the screen shots  to this PR top comment.

madolson:
@stockholmux @ranshid Do either of you two want to look it over? I didn't really look into the JS code that closely.

templates/valkey-try-me.html
```diff
@@ -0,0 +1,232 @@
...
+                memory_size: 512 * 1024 * 1024,
+                bios: { url: "https://download.valkey.io/try-me-valkey/vos/v86/bios/seabios.bin" },
+                filesystem: {
+                    baseurl: "https://download.valkey.io/try-me-valkey/8.1.0/fs/alpine-rootfs-flat",
+                    basefs: "https://download.valkey.io/try-me-valkey/8.1.0/fs/alpine-fs.json",
+                },
+                autostart: true,
+                bzimage_initrd_from_filesystem: true,
+                cmdline: "rw root=host9p rootfstype=9p rootflags=trans=virtio,cache=loose modules=virtio_pci tsc=reliable",
+                initial_state: { url: imgUrl },
```
SuperMaxusa, line 147: You can use `zstd` instead of `gzip` for the state file, v86 supports this out of the box.

PR #257, opened 2025-05-08
Adding an ACL section for each command on their command page

zackcam:
### Description

This will add a section underneath the since on command pages to include ACL categories. This will add an @ in front of each category a comma inbetween and will set as lowercase.
#### Single acl category:
image

#### Works with module commands as well:
image

As mentioned in the issue, non module commands only have their data type in the ACL categories. If this is a blocking issue I can see about adding all acl categories to the command pages for commands

### Issues Resolved

This is based off of https://github.com/valkey-io/valkey-doc/issues/272 

### Check List

### DIFF
--- a/templates/command-page.html
+++ b/templates/command-page.html
@@ -78,6 +78,16 @@
         <dd>{{ command_data_obj.since }}</dd>
     <dl>
     {% endif %}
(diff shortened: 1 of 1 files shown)


pnbrown (approved):
LGTM

PR #264, opened 2025-05-22
feat: add community page with navigation updates

phiro56:
- Create community page template and content
- Update navigation menu to match new design and include Community page link

screencapture-127-0-0-1-1111-community-2025-05-22-13_20_39


templates/default.html
```diff
@@ -24,23 +24,11 @@
       ></div>
       <nav role="navigation" aria-label="Main">
         <a role="menuitem" href="/download/">Download</a>
-        <div class="has-submenu">
-          <a role="menuitem" href="/docs/">Documentation</a>
-            <div class="submenu">
-              <a role="menuitem" href="/commands/">Command Reference</a>
-              <a role="menuitem" href="/topics/">Documentation by Topic</a>
-              <a role="menuitem" href="/clients/">Client Libraries</a>
```
madolson, line 32: Client libraries is getting dropped. I know some stuff has changed since the original design, I think let's keep it simple and keep the documentation drop down and have command reference, documentation, and client libraries under it. Everything else can stay as is.

PR #285, opened 2025-06-27
Remove duplicate communnity and participants

madolson:
### Description

It seems that when we merged try-my valkey we had a conflict where we duplicated some banner fields. This is just removing them.

image

 
### Issues Resolved

No

### Check List


stockholmux (approved):
LGTM.

Although I don't like how this is hard coded. It should be driven by a config file and template loop so this won't happen again.

madolson:
> Although I don't like how this is hard coded. It should be driven by a config file and template loop so this won't happen again.

Will make an issue.

PR #293, opened 2025-07-04
Extend left sidebar background to full width

phiro56:
Extend left sidebar background to full width

Screenshot 2025-07-03 at 11 17 18 PM


stockholmux (approved):
LGTM

PR #294, opened 2025-07-06
feat(ui): Add FAQ page with styled sections and category navigation

phiro56:
## Update FAQ UI
feat(ui): Add FAQ page with styled sections and category navigation

- Move FAQ content from external source to local repository
- Create new FAQ template with left sidebar navigation
- Add styled FAQ sections with collapsible content
- Implement category-based navigation links
- Add review notice banner for pages under review

Screenshot 2025-07-24 at 6 12 51 PM


stockholmux:
Looks like we're going to port the FAQ from docs to the website repo.

phiro56:
@stockholmux latest changes include:

1. Keep content/topics/faq.md under version control in this repo. FAQ content is now managed locally, tracked by git, and can be edited directly in your main website repo. No more dependency on the remote valkey-doc repo for FAQ content.

2. Protect Local FAQ from Being Overwritten. By adding a check in build/init-topics-and-clients.sh to skip faq.md, we ensure that the build script will never overwrite your local FAQ when it creates or updates other topic stubs.

content/topics/faq.md
```diff
@@ -0,0 +1,175 @@
++++
+title = "FAQ"
+template = "faq.html"
++++
+
+<div class="faq-section">
```
stockholmux, line 6: Putting HTML in a markdown file makes it hard to edit and prone to mistakes.

templates/faq.html
```diff
@@ -0,0 +1,79 @@
...
+            <div class="edit-box">
+                <span>Found an Issue?</span>
+                <a href="https://github.com/valkey-io/valkey-io.github.io/edit/main/content/topics/faq.md" class="edit-link" target="_blank">
+                    Edit this page on GitHub
+                </a>
+            </div>
+        </aside>
+
+        <main class="container">
+            {% if config.extra.review_list is containing(page.path) %}
```
stockholmux, line 31: I don't think we need the review list bits. This appears to be duplicative from the `docs-page.html` file which only exists as a function of importing a bunch of dubious content. This isn't imported so I think we can drop all that.

templates/faq.html
```diff
@@ -0,0 +1,79 @@
...
+                <div class="review-notice">
+                    <p><strong>This page is under review.</strong> The page is likely incorrect, contains invalid links, and/or needs technical review. 
+                    In the future it may change substantially or be removed entirely.</p>
+                </div>
+            {% endif %}
+
+            <div class="faq-content">
+                {% if config.extra.publish_hold is containing(page.path) %}
```
stockholmux, line 39: We can drop this block

build/init-topics-and-clients.sh
```diff
@@ -40,6 +40,11 @@ do
     base=${fname##*/}
     topic=${base%.*}
 
+    # Skip faq.md so it is never overwritten
```
stockholmux, line 43: I don't think this is the correct approach - this basically bypasses the usefulness of the SSG.

README.md
```diff
@@ -61,6 +61,8 @@ Point your browser at `http://127.0.0.1:1111/topics/` and you should see the ful
 All files created in this process are ignored by git.
 Commit your changes to your local copy of `valkey-io/valkey-doc`.
 
+**Note:** `content/topics/faq.md` is managed locally in this repo and will not be overwritten by the build script.
```
stockholmux, line 64: We should change this approach.

stockholmux (changes requested):
This doesn't seem like an approach that will be easy to update or use.

Can we use a [Zola section](https://www.getzola.org/documentation/content/section/) and [page](https://www.getzola.org/documentation/content/page/) with `render = false`? 

Ideally, each FAQ item is built from it's own markdown file: the question would be a front matter item, and the answer would be the body of the file. The FAQ categories would be a [taxonomy](https://www.getzola.org/documentation/content/taxonomies/).

This would make it easy to update and add more. Once the FAQ is ported over, we can just get rid of the file on `valkey-doc`

PR #330, opened 2025-08-07
generate once, load many

### LINKED ISSUE #321: [BUG] Build time explosion due to command page sidebar
**Describe the bug**
Related to #319 and #320

I've noticed that the build times have shot up after #299 to north of around 50 seconds - which is bonkers. 

It appears this is due to the inclusion of the command page sidebar. 

Given the repetitious nature of this page element, this should be abstracted into a macro and/or cached during build. At time of writing there is 416 pages rendered in this section, each one has to open 416 + 4 (json files that index all the commands). That means that generating this sidebar adds 174,720 JSON file opens and parses.

**To Reproduce**

Run `zola serve` note CPU usage and the build time.

On my machine the build time is between 45 and 76 seconds and typically pegs the CPU to over 200%. 

**Expected behavior**

The site should build in less than 10 seconds.

**Screenshots**

n/a
**Host/Environment (please complete the following information):**

macOS

**Additional context**

n/a

stockholmux:
### Description

When the side bar was added to the command page it caused a massive increase in build times (see #321). This was caused by each command page needing to pull every command JSON file into to generate the list.

This PR takes a different approach, it adds a page (`/command-list/`), that is generated once, which contains the all the commands formatted as list items. Then, on each command page, it pulls that content in with some simple javascript. 

Full build times were previously taking 40-70 seconds. After this change, the build times are taking approximately 1.2 seconds.

```
Building site...
Checking all internal links with anchors.
> Successfully checked 0 internal link(s) with anchors.
-> Creating 564 pages (0 orphan) and 8 sections
Warning: Highlight language plaintext not found
Done in 1.2s.
```
 
### Issues Resolved

#321

### Check List

### DIFF
--- a/content/command-list/_index.md
+++ b/content/command-list/_index.md
@@ -1,0 +1,5 @@
++++
+title = "Command List"
+template = "command-list.html"
+page_template = "blog-page.html"
++++
--- a/templates/command-list.html
+++ b/templates/command-list.html
@@ -1,0 +1,35 @@
+{% import "macros/command.html" as commands %}
+
+{% set commands_entries = [] %}
+{% set commands_section = get_section(path="commands/_index.md") %}
+{% for page in commands_section.pages %}
+
+    {% for json_path in [
+        commands::command_json_path(slug=page.slug),
+        commands::command_bloom_json_path(slug=page.slug),
+        commands::command_json_json_path(slug=page.slug),
+        commands::command_search_json_path(slug=page.slug)
+    ] %}
+        {% set command_data = load_data(path= json_path, required= false) %}
+        {% if command_data %}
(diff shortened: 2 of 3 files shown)


stockholmux:
I'll also add that this is vastly better for SEO which will ignore the hundreds of links now being included by JS.

madolson (approved):
I think I would prefer it to be built statically, but I don't feel strongly.

stockholmux:
@madolson Yeah, I think that's the idiomatic way to think about this site, but this is a case where a dynamic inclusion on the front-end is the right choice: it rarely changes (so cached by the browser), it takes almost no logic to include or parse, and it's 45x faster to build plus the SEO bit, usually a downside of dynamic inclusion, but 400+ links in a sidebar will just confuse a crawler.

PR #380, opened 2025-10-05
feat: add go module redirect for k8s-operator

dmolik:
### Description

the Golang import path for the new valkey-operator is valkey.io/valkey-operator this patch wires up a redirect to use the patch  but also keep the code within GitHub.
 
### Check List

### DIFF
--- a/content/valkey-operator/_index.md
+++ b/content/valkey-operator/_index.md
@@ -1,0 +1,4 @@
++++
+template = "operator.html"
+title = "Valkey-Operator"
++++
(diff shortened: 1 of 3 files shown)


stockholmux:
@dmolik So it looks like you don't actually need any of the rest of the template and this page wouldn't really be visited by humans?

If so, I think a more elegant way of doing this would be to create a standalone go-forwarding template that doesn't rely on the existing templates but has the two meta tags you need for Golang.

As-is would work but we're trying to gently optimize for a nice quick build process - this conditional would be evaluated on every page of the website - we can handle it today but too many of these tech debt over time.

stockholmux:
Closing this as the [operator has opted to go with GitHub directly](https://github.com/valkey-io/valkey-operator/pull/316). Feel free to re-open if needed.

PR #400, opened 2025-10-22
Fix bash code rendering in 1 billion RPS blog post

ei-grad:
### Description

Remove unnecessary Tera template escaping from bash code blocks. Code fences
are not processed by Tera templating, so the double braces were rendering
literally as "{{" instead of "{", breaking the bash syntax for awk blocks
and variable expansions.
 
### Check List


stockholmux (approved):
LGTM

hpatro (approved):
Thanks!

sarthakaggarwal97:
Thank you @ei-grad! We were running inside Python f-string, that's how the extra brackets crept in.

Issue #416, opened 2025-11-20
Update Valkey GLIDE client links.


Aryex:
## Descriptions
With the new Valkey GLIDE documentations going into public preview on November 21st, 2025, the GLIDE client links on the site should be udpated.

## Changes
As previously [discussed](https://github.com/valkey-io/valkey-io.github.io/issues/383), we would like the new GLIDE site to be added to the official [valkey.io](http://valkey.io/) client list. Currently, GLIDE clients are on the list but their documentation link will need to be updated to the new site at valkey.io/valkey-glide-docs

We also would like to propose using a separate section for GLIDE instead. This is not a blocker for us to publish the preview but do let us know your thoughts.

PR #418, opened 2025-11-21
Support website links in client list and fix template variable shadowing

Aryex:
## Overview
This PR is a proposal to add a website link to the client list page to support linking the new Valkey GLIDE website. 

It also fixes a bug in the template file.

## Issue
The client page only display repo links and does not support website link.

### Adding Website Links

The template now displays a link to the client's website, if one is provided in the client's data file. This is handled by conditionally rendering the link only when a non-empty website URL is available.

Here's what the python section looks like:
image

### Fixes Variable Shadowing

Resolves an issue where a `client` variable was being shadowed within a loop, which could lead to unpredictable behavior. The shadowing variable has been renamed to `first_client`.

### Related issues
https://github.com/valkey-io/valkey-io.github.io/issues/416

### DIFF
--- a/templates/client-list.html
+++ b/templates/client-list.html
@@ -37,6 +37,9 @@
                             <h3 id="{{ client.name | slugify }}">{{ client.name }}</h3>
                             <ul>
                                 <li><strong>Repo:</strong> <a href="{{ client.repo }}">{{ client.name }}</a></li>
+                                {% if client.website %}
+                                    <li><strong>Website:</strong> <a href="{{ client.website }}">{{ client.website }}</a></li>
+                                {% endif %} 
                                 <li>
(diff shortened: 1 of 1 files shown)


stockholmux (approved):
LGTM. @Aryex Did the 'variable shadowing' actually cause problems? My understanding is the tera templates had short term contexts unless explicitly global.

Aryex:
> LGTM. @Aryex Did the 'variable shadowing' actually cause problems? My understanding is the tera templates had short term contexts unless explicitly global.

It seems so. For clients without a website, it was still rendering the first website `Website: https://glide.valkey.io/languages/python/` from the first client (Glide Python) even though it nothing.

When I tried renaming the clients it fixed it.

PR #446, opened 2026-01-29
Releases pages link and (optionally) forward to latest versions

### LINKED ISSUE #428: [BUG] No way to link to a latest patch
**Describe the bug**

There are cases where it makes sense to link to a latest patch (e.g. 9.1.x or 9.0.x). Right now we can only link to specific releases (e.g. 9.1.0 or 9.0.1). So there needs to be a way to link to the latest patch - e.g `/download/releases/v9-0-x/` that forwards to `/download/releases/v9-0-0/` if only 9.0.0 exists or `/download/releases/v9-0-1/`if 9.0.1 exists.

stockholmux:
### Description

Release pages provide a stable URL to point to a version of the software. For example, `/download/releases/v7-2-5/` will always have things like the GitHub tag, release date, and artifact links.

In the past, there has been situations where we've made a major release, found a bug, and did a fast-followed with a patch release to fix it. However, the blog post for that release still points to the release page for the original release, obscuring a bug fix. This has two issues:

1. Release pages don't hint at a newer release being available.
2. There isn't a way to link to a newer release in the same line (patch/minor) (e.g. #428) or to the latest major release

This PR adds a few things around this problem:

#### Release pages have an `Available updates` section.

On each release page, there is section that advises about updates/upgrades, linking to the release pages for each.

Screenshot 2026-01-29 at 7 50 46 AM

The script only shows the relevant updates for each version. As an example, 7.2.11 doesn't show 8.x releases, only the latest major:

Screenshot 2026-01-29 at 7 53 46 AM

On pages which don't have any updates, the section isn't present.

#### URL hash and forwards

To directly fix #428, this PR adds an optional hash tag to the URL. If you add `#latest_patch`, `#latest_minor` or `#latest_major` to any release page (e.g.  `/download/releases/v7-2-5/`) it will do a javascript forward to the respective latest version. For example:
- `/download/releases/v7-2-5/#latest_patch` ->  `/download/releases/v7-2-11/`
- `/download/releases/v8-0-0/#latest_minor` -> `/download/releases/v8-1-5/`
- `/download/releases/v8-1-5/#latest_major` -> `/download/releases/v9-0-1/`
 
If there isn't new latest for the different hash tags, it's a noop: `/download/releases/v9-0-1/#latest_patch` does nothing. So, for release blogs, we can safely use URLs like `download/releases/v9-0-0/#latest_patch` and users will automatically get the latest patch version. No more obscured releases!

Release candidates are excluded from this logic. 

#### Bug fixes

In writing this feature, I noticed a bug where the multiple releases fragments on the same page (such as on the homepage or `/downloads/`) misreported the release date for all releases as the first listed. 🤦‍♂️  This is a subtle bug because this _often_ but not always the case. 

I additionally tested what 10.0.0 would look like and discovered that, in some context 10.0.0 appeared in the wrong order. Now it's explicitly uses date sorting - the most recent & version numbers are used and 10.0.0 will come out on top.

#### Refactoring

I moved the release fragment from being a zola include to being a macro. This isolates the scope better to prevent the release date bug as above and actually makes the code a little cleaner. 

### Issues Resolved

#428 

### Check List

### DIFF
--- a/templates/download.html
+++ b/templates/download.html
@@ -1,4 +1,5 @@
 {% extends "right-aside.html" %}
+{% import "macros/release.html" as release_macros %}
 
 {% block subhead_content%}
 <h1 class="page-title">Download Latest</h1>
@@ -31,7 +32,7 @@ The <a href="./releases/">releases</a> page contains links to download all curre
 
 
 <h3> Version {{major}}.x.x </h3>
-{% include "includes/release.html" %}
+{{ release_macros::release_info(date= most_recent_release_page[0].date, release= most_recent_release_page[0].extra ) }}
 
 <h2> Latest supported version of past releases</h2>
 
@@ -42,7 +43,7 @@ The <a href="./releases/">releases</a> page contains links to download all curre
         {% set_global major = older_major %}
         {% set release = older_release.extra %}
         <h3> Version {{major}}.x.x </h3>
-        {% include "includes/release.html" %}
+        {{ release_macros::release_info(date= older_release.date, release= older_release.extra ) }}
     {% endif %}
 {% endfor %}
 
--- a/templates/index.html
+++ b/templates/index.html
@@ -1,4 +1,5 @@
 {%- extends "default.html" -%}
+{% import "macros/release.html" as release_macros %}
 
 {%- block content -%}
 {% if section.extra and section.extra.hero %}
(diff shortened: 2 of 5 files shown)


madolson (commented):
I should post this comment since I've just been staring at this. 

Having a reference link that does just an update seems wrong. I was always assuming we would just have a fixed page for each major.minor version, that includes the links to all the old patch versions probably made more sense.

I don't know if I have a strong opinion though. I know we've also talked about having metadata about what were the flagship features in each release.

stockholmux:
closing as obsolete due to #662

PR #488, opened 2026-04-02
Add blog post type tags with filtering and build-time validation

madolson:
### Description

Adds category tags to all blog posts with a filterable dropdown on the blog listing page and badge display on individual post pages.

- Every blog post now has a `type` field in its `[extra]` front matter
- Blog listing page displays a type badge per post and a dropdown to filter by type
- Individual post pages show the type badge in the metadata header
- URL hash is preserved correctly when clearing the type filter (`location.pathname + location.search` instead of `location.pathname` alone), so existing query parameters are not dropped

The 5 current blog types:
- `Technical Deep Dive` — In-depth technical content
- `Announcements` — Releases, new features, project news
- `Community Highlight` — Community stories, year-in-review, ecosystem
- `How-to` — Tutorials and guides
- `Events` — Event recaps and announcements

### Check List


stockholmux (changes requested):
This is a really strange way to try to achieve what I think you're going after.

Use the built-in [Taxonomies](https://www.getzola.org/documentation/content/taxonomies/). It's way more flexible and doesn't require custom logic every time we do a section loop. Taxonomies also gives feeds and doesn't break localizations (not used currently, but requested).

stockholmux (approved):
I'm broadly okay with this as a start.

We should create a `taxonomy_list.html` template to drive real HTML rendered pages instead of the drop down selector, as well as turn on taxonomy feeds but I don't think either need to block.

PR #514, opened 2026-04-24
Add workflow to auto-assign technical-blog-reviewers

madolson:
### Description

Adds a GitHub Actions workflow that automatically requests review from `@valkey-io/technical-blog-reviewers` when a PR adds a new blog post with `blog_type` of "Technical Deep Dive" or "Announcements".

This also updates code owners a bit. Removes some files that don't need Madelyn/Kyle's review while keeping kyle assigned to templates. All content get's content creators added (we can change the name, but that was the historical group that does blogs/events)

**How it works:**
1. Triggers on `pull_request_target` (opened/synchronize) when files under `content/blog/**` change
2. Uses the GitHub API to detect newly added `.md` files (ignores modifications to existing posts)
3. Fetches each new file's content and parses the TOML front matter for `blog_type`
4. Requests review from `technical-blog-reviewers` only if a matching type is found

**Design decisions:**
- Uses `pull_request_target` so it works for PRs from forks (most external contributors)
- Reads file content via the GitHub API only — no checkout of PR code (safe for `pull_request_target`)
- Regex-based front matter parsing — the `blog_type` format is consistent across all 44 existing blog posts
- Gracefully handles missing `blog_type` and API errors

### Check List

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,2 @@
-*   @AMoo-Miki @madolson @stockholmux
+*              @AMoo-Miki @madolson @stockholmux
+/templates/    @stockholmux
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,3 +1,6 @@
 *              @madolson @stockholmux
-/content/      @valkey-io/content-creators
+/content/blog/       @valkey-io/content-creators
+/content/events/     @valkey-io/content-creators
+/content/community/  @valkey-io/content-creators
+/content/authors/    @valkey-io/content-creators
 /templates/    @stockholmux
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,2 +1,3 @@
 *              @AMoo-Miki @madolson @stockholmux
+/content/      @valkey-io/content-creators
 /templates/    @stockholmux
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,3 +1,3 @@
-*              @AMoo-Miki @madolson @stockholmux
+*              @madolson @stockholmux
 /content/      @valkey-io/content-creators
 /templates/    @stockholmux
--- a/.github/workflows/assign-blog-reviewers.yml
+++ b/.github/workflows/assign-blog-reviewers.yml
@@ -39,7 +39,6 @@ jobs:
 
             const technicalTypes = ['Technical Deep Dive', 'Announcements'];
             let needsTechnical = false;
-            let needsContent = false;
 
             for (const file of newBlogFiles) {
               // Fetch file content from the PR head SHA via the API (safe for fork PRs)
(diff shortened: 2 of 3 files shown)


.github/CODEOWNERS
```diff
@@ -1 +1,3 @@
-*   @AMoo-Miki @madolson @stockholmux
\ No newline at end of file
+*              @madolson @stockholmux
+/content/      @valkey-io/content-creators
```
stockholmux: so `/content/` is more than just blog posts.  Should it be scoped to `/content/blog/`?

.github/workflows/assign-blog-reviewers.yml
```diff
@@ -0,0 +1,84 @@
...
+                owner: context.payload.pull_request.head.repo.owner.login,
+                repo: context.payload.pull_request.head.repo.name,
+                path: file.filename,
+                ref: head_sha,
+              });
+
+              const content = Buffer.from(data.content, 'base64').toString('utf-8');
+
+              // Extract blog_type from TOML front matter (between +++ fences)
+              const match = content.match(/blog_type\s*=\s*\["([^"]+)"\]/);
```
stockholmux: I am suspicious of this.

Frontmatter can be in either TOML or YAML. How would that work? Also in, theory, a single post can have multiple taxonomies - not that we really use it that way.

stockholmux (commented):
Broadly looks okay with the approach but I think the PR won't work in all cases.

.github/CODEOWNERS
madolson: It's also events and some other minor things, I wanted that to also be inclusive of that group and just sort of ask all of them to own it. Happy to discuss alterantives.

.github/workflows/assign-blog-reviewers.yml
madolson: It can be YAML? Why are we not using YAML, it's so much better than TOML :).

In either case. Let me change it to throw an error, I guess we want it to error out anyways.

stockholmux: You made me doubt myself, but indeed: [YAML is supported but not encouraged in Zola] (https://github.com/getzola/zola/blob/master/docs/content/documentation/content/section.md?plain=1#L42). We really shouldn't encourage YAML at all. [Think of Norway](https://www.bram.us/2022/01/11/yaml-the-norway-problem/)!

.github/CODEOWNERS
stockholmux: I might scope it to be more specific to areas where content happens. The non-content parts of `/content/` (ironic) could break things like downloads, docs, clients (that last one is a pain... we should look at reworking that one generally).

stockholmux (commented):
Generally aligned.

.github/CODEOWNERS
madolson: Ok, the best list I came up with was community, events, blogs, and authors.

stockholmux (approved):
LGTM

PR #534, opened 2026-04-29
feat: add category filtering to commands page sidebar

### LINKED ISSUE #525: Command reference category links should filter, not just scroll
## Problem

On the [Command Reference](https://valkey.io/commands/) page, clicking a category in the sidebar (e.g., "Search" at `/commands/#search`) only scrolls the browser to that group heading. All other command groups remain visible on the page.

With 200+ commands across 20+ groups, this makes the category navigation feel broken and incomplete. Users expect clicking "Search" to show only search commands, similar to how the search box filters results in real time.

## Current behavior

1. User clicks "Search" in sidebar
2. Page scrolls to the Search group heading
3. All other groups (String, List, Hash, etc.) remain visible above and below
4. No visual indicator of which category is active in the sidebar

## Expected behavior

1. User clicks "Search" in sidebar
2. Only Search commands are shown (other groups hidden)
3. Sidebar highlights "Search" as the active category
4. URL updates to `/commands/#search` (shareable)
5. Clicking "All" or clearing the filter shows everything again

## Suggested implementation

Add a small JavaScript handler that:

1. Listens for `hashchange` events and checks `location.hash` on page load
2. When a group hash is present, hides all `.command-group` divs except the matching one (same pattern the existing search box uses)
3. Adds an `.active` class to the matching sidebar link
4. Clears the filter when hash is removed or a "Show All" link is clicked

The existing `searchCommands()` function already implements show/hide logic for `.command-group` and `.command-entry` elements, so this would follow the same pattern.

kkonaparthi1:
Fixes #525

## Summary
Clicking a command category in the sidebar now filters the page to show only that category's commands, instead of just scrolling to the heading while all 200+ commands remain visible.

## Changes
- Added `hashchange` event listener that hides all command groups except the selected one
- Added active state styling on the selected sidebar link
- Added "Show All" option to restore the full command list
- Clears search box when filtering by category
- Filters on initial page load if URL contains a hash

## Approach
Follows the same show/hide pattern already used by the existing `searchCommands()` function. ~30 lines of JavaScript + 5 lines of SCSS.

### DIFF
--- a/templates/commands.html
+++ b/templates/commands.html
@@ -92,11 +92,70 @@ function searchCommands() {
     // Show/hide no results message based on search results
     noResultsMessage.style.display = totalVisible === 0 ? "block" : "none";
 }
+
+function filterByCategory() {
+    var hash = window.location.hash.substring(1);
+    var groups = document.querySelectorAll(".command-group");
+    var sidebarLinks = document.querySelectorAll(".category-link");
+    var noResultsMessage = document.getElementById("no-results-message");
+
+    // Clear search box when filtering by category
+    var searchBox = document.getElementById("search-box");
+    if (searchBox) searchBox.value = "";
+
+    // Update active state on sidebar links
+    sidebarLinks.forEach(function (link) {
+        link.classList.remove("active");
+        if (link.getAttribute("href") === "#" + hash) {
+            link.classList.add("active");
+        }
+    });
+
+    // If no hash or hash is "top", show all groups
+    if (!hash || hash === "top") {
+        groups.forEach(function (group) {
+            group.style.display = "";
+            group.querySelectorAll(".command-entry").forEach(function (item) {
+                item.style.display = "";
+            });
+        });
+        noResultsMessage.style.display = "none";
+        sidebarLinks.forEach(function (link) { link.classList.remove("active"); });
+        var showAllLink = document.querySelector(".show-all-link");
+        if (showAllLink) showAllLink.classList.add("active");
+        return;
+    }
+
+    // Show only the matching group
+    var found = false;
+    groups.forEach(function (group) {
+        var heading = group.querySelector("h2");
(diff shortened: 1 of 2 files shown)


templates/commands.html
```diff
@@ -92,22 +92,82 @@ <h2 id="{{command_group_description_name}}">{{ group_descriptions[command_group_
...
+    // Filter when hash changes (sidebar link clicked)
+    window.addEventListener("hashchange", filterByCategory);
 });
 </script>
 {% endblock main_content %}
 
 {% block related_content %}
 <h2 id="top">Command Categories</h2>
 <ul>
+<li><a href="#top" class="show-all-link active" onclick="window.location.hash='top';">Show All</a></li>
```
stockholmux, line 166: The 'Show all' button is active when you don't have any filtering, which is kind of confusing.

Here I'm not filtering, so I'd expect "Show all" to be hidden.
Image

templates/commands.html
```diff
@@ -92,22 +92,82 @@ <h2 id="{{command_group_description_name}}">{{ group_descriptions[command_group_
...
 {% if group_descriptions %}
 {% for command_group_name, group_description in group_descriptions %}
     {% set replaced_group_id = command_group_name | replace(from="-", to="_") %}
-    {% if grouped[replaced_group_id]  %}<li>  <a href="#{{ command_group_name }}">{{ group_description.display }}</a></li>{% endif %}
+    {% if grouped[replaced_group_id]  %}<li>  <a href="#{{ command_group_name }}" class="category-link">{{ group_description.display }}</a></li>{% endif %}
```
stockholmux, line 170: It would probably make sense to have some sort of indication that you're filtering above the search bar. 

For example: "Show only **Search** commands."

Image

stockholmux (changes requested):
Really nice work! A few little UX tweaks needed to make it easily usable.

PR #587, opened 2026-06-25
Update 2026 events calendar

mtuteja-git:
### Description

- Add 2026 events to the Valkey events calendar
- Add Meetup / Hackathon event type
- Add ValkeyConf 2026 and updated event links
 
### Check List

Screenshot 2026-06-24 at 10 52 04 PM
Screenshot 2026-06-24 at 10 52 13 PM
Screenshot 2026-06-24 at 10 52 22 PM
Screenshot 2026-06-24 at 10 52 28 PM
Screenshot 2026-06-24 at 10 52 37 PM


pnbrown:
Is there anything we can do about the clipping at the bottom of the list under the calendar?
Screenshot 2026-06-25 at 14 54 52

pnbrown:
It feels like the footer takes over too early. The page should be longer? The scroll area is too small

mtuteja-git:
@pnbrown it is not clipping, it is cos of the screenshot. 

Regarding the scroll area, it is proportionate to the calendar - do you have a recommendation for the proportions?

Screenshot 2026-06-25 at 12 02 08 AM

pnbrown:
Responsive to fit the month?

mtuteja-git:
yes.

templates/events.html
```diff
@@ -105,7 +177,7 @@ <h1 class="page-title">Events</h1>
       if (hasEvents) {
         var labels = '';
         hasEvents.forEach(function(e) {
-          labels += '<a href="' + e.url + '" class="event-label ' + e.type + '">' + e.title + '</a>';
+          labels += '<a href="' + e.url + '" class="event-label ' + e.type + '" title="' + e.title + '">' + e.title + '</a>';
```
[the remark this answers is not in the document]
rlunar: @mtuteja-git please apply the suggested fix by greptile

templates/events.html
```diff
@@ -139,21 +211,20 @@ <h1 class="page-title">Events</h1>
               String(nowLocal.getMonth() + 1).padStart(2, '0') + '-' +
               String(nowLocal.getDate()).padStart(2, '0');
   var currentMonth = today.substring(0, 7);
-  var thisMonthUpcoming = events.filter(function(e) { return e.date.substring(0, 7) === currentMonth && e.date >= today; })
+  var thisMonthUpcoming = events.filter(function(e) { return e.date.substring(0, 7) === currentMonth && eventEndDate(e) >= today; })
                         .sort(function(a, b) { return a.date.localeCompare(b.date); });
-  var upcoming = events.filter(function(e) { return e.date > today && e.date.substring(0, 7) !== currentMonth; })
+  var upcoming = events.filter(function(e) { return eventEndDate(e) >= today && e.date.substring(0, 7) !== currentMonth; })
                        .sort(function(a, b) { return a.date.localeCompare(b.date); });
```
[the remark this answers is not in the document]
rlunar: @mtuteja-git please apply the fix suggested by coderabbit to fix the events that cross months like Laracon India

rlunar (commented):
I agree with this PR overall; there are some edits suggested by CodeRabbit and Greptile that need to be addressed.

What do we consider a Valkey event? for example we also did CNCF Mexico City Meetup on June 9th, does that count?

Should we have a naming convention? for example KCD's are prefixed properly, but MeetUps are not; should we make ALL events to follow a taxonomy?

mtuteja-git:
> I agree with this PR overall; there are some edits suggested by CodeRabbit and Greptile that need to be addressed.
> 
> What do we consider a Valkey event? for example we also did CNCF Mexico City Meetup on June 9th, does that count?
> 
> Should we have a naming convention? for example KCD's are prefixed properly, but MeetUps are not; should we make ALL events to follow a taxonomy?

Addressed the changes suggested by CodeRabbit and Greptile.

A CNCF Mexico City Meetup is not a 1P event or a meetup/hackathon if it was not a Valkey meetup but if we had presence there it can qualify as a 3P event (third party)

I like that idea, so far it was been Valkey Meetup *City with an additional line for SEO/relevance, which I am not super against.

mtuteja-git:
@valkey-io/content-creators can you please review so we can add events to the website

crystalphamLF:
Sorry @mtuteja-git - it wont let me squash and merge.

templates/events.html
```diff
@@ -48,16 +50,74 @@ <h1 class="page-title">Events</h1>
...
+  function eventMonthKeys(e) {
+    var keys = [];
+    eventDateKeys(e).forEach(function(dayKey) {
+      var monthKey = dayKey.substring(0, 7);
+      if (keys.indexOf(monthKey) === -1) keys.push(monthKey);
+    });
+    return keys;
+  }
+
+  function formatDateRange(startKey, endKey) {
```
stockholmux: This function is odd. It assume Month/Day/Year which is only really common in the USA.

You can use [`.toLocaleDateString()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString#try_it) to auto format to whatever the browser is operating in.

tests/events-2026.test.mjs
```diff
@@ -0,0 +1,161 @@
+import assert from "node:assert/strict";
```
stockholmux: While I generally applaud the use of tests we haven't agreed to this testing framework nor have we described how to do tests anywhere. I think we should drop this bit.

tests/events-2026.test.mjs
```diff
@@ -0,0 +1,161 @@
...
+import { readFileSync, existsSync } from "node:fs";
+import { join } from "node:path";
+
+const repoRoot = process.cwd();
+
+const expectedEvents = [
```
stockholmux: I also think this will break when we add events :)

templates/events.html
```diff
@@ -93,7 +166,14 @@ <h1 class="page-title">Events</h1>
 
     // Build event lookup by day
     var byDay = {};
-    evts.forEach(function(e) { var d = Number(e.date.substring(8, 10)); if (!byDay[d]) byDay[d] = []; byDay[d].push(e); });
+    evts.forEach(function(e) {
+      eventDateKeys(e).forEach(function(dayKey) {
+        if (dayKey.substring(0, 7) !== key) return;
+        var d = Number(dayKey.substring(8, 10));
```
stockholmux: You're doing a ton of processing of dates in this script. I worry that it might be a little fragile and it's kind of unnecessary considering you can just use Zola to pre-generate all this as part of `events` for loop.

stockholmux (changes requested):
@mtuteja-git No problems with the events content (yay!), but there is some Javascript here that I would question. I've been looking at it for a while and there is places where you format the date, then later pull that formatting apart and reformat it. 

The end result is probably not an issue but date objects tend to have corner cases that don't show up until later and/or are interpreted by slightly differently across browsers and locales. We can probably shift most of this up to Zola and know it's repeatable forever.

tests/events-2026.test.mjs
madolson: +1, I don't think we should add tests for things where the real test is verifying the external functionality.

templates/events.html
mtuteja-git: I removed the manual date formatting and am using .toLocaleDateString() for the displayed date labels - please review and let me know if this is what you were expecting.

tests/events-2026.test.mjs
mtuteja-git: Thank you, dropped.

tests/events-2026.test.mjs
mtuteja-git: Yes, removed this test file.

templates/events.html
mtuteja-git: Removed the 'eventDateKeys' / 'eventMonthKeys' / 'formatDateRange' helper layer and simplified the calendar logic to use ISO date range checks. The displayed dates now use 'toLocaleDateString()' - please review when you can and let me know if that looks correct to you.

templates/events.html
```diff
@@ -50,72 +50,84 @@ <h1 class="page-title">Events</h1>
...
+    return new Date(Number(parts[0]), Number(parts[1]) - 1, Number(parts[2]));
   }
 
-  function eventMonthKeys(e) {
-    var keys = [];
-    eventDateKeys(e).forEach(function(dayKey) {
-      var monthKey = dayKey.substring(0, 7);
-      if (keys.indexOf(monthKey) === -1) keys.push(monthKey);
+  function displayDate(dateKey) {
+    return localDateFromKey(dateKey).toLocaleDateString(undefined, {
```
stockholmux: I think this part is incorrect.

`localDateFromKey` function is dubious as it's just create a date object in a really weird way. I think just `new Date(dateKey)` would accomplish the same thing. Then the `toLocaleDateString` is always create numeric-long-numeric which may not be locale specific at all.

stockholmux: (Also, see code rabbit later.. it's saying partially the same thing).

stockholmux (dismissed):
Locale and date stuff is still super weird. `parseDateKey` was removed and then the same code was added as `localDateFromKey` - I'm not entirely sure any of it is needed.

templates/events.html
```diff
@@ -48,6 +49,49 @@ <h1 class="page-title">Events</h1>
...
+    return {
+      title: e.title,
+      safeTitle: escapeHtml(e.title),
+      date: e.date,
+      eventClass: type,
+      typeLabel: eventTypeLabels[type],
+      location: e.location,
+      safeLocation: escapeHtml(e.location),
+      safeLocationUrl: safeHref(e.locationUrl),
+      safeUrl: safeHref(e.url)
```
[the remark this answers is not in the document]
madolson: That is what we want for now.

templates/events.html
```diff
@@ -48,6 +49,49 @@ <h1 class="page-title">Events</h1>
...
+  };
+
+  function escapeHtml(value) {
+    return String(value)
+      .replace(/&/g, '&amp;')
+      .replace(/"/g, '&quot;')
+      .replace(/'/g, '&#39;')
+      .replace(/</g, '&lt;')
+      .replace(/>/g, '&gt;');
+  }
```
madolson, line 58: Minor: the root cause of needing this machinery is that the Tera loop emits values directly into JS string literals with `| safe` (lines 42-47). If you emit them with `json_encode` instead, Zola produces correctly quoted and escaped JS literals, and the whole class of quote-injection bugs disappears at build time:

```
title: {{ page.title | json_encode() | safe }},
```

Note `json_encode` adds the surrounding quotes, so drop the manual ones. `escapeHtml` is still needed where we build innerHTML, but this moves the data-transport problem up to Zola, which is what @stockholmux was suggesting.

templates/events.html
```diff
@@ -48,6 +49,49 @@ <h1 class="page-title">Events</h1>
...
+      safeTitle: escapeHtml(e.title),
+      date: e.date,
+      eventClass: type,
+      typeLabel: eventTypeLabels[type],
+      location: e.location,
+      safeLocation: escapeHtml(e.location),
+      safeLocationUrl: safeHref(e.locationUrl),
+      safeUrl: safeHref(e.url)
+    };
+  });
```
madolson, line 80: Minor: this map creates parallel raw and safe copies of each field (`title`/`safeTitle`, `location`/`safeLocation`), but the raw ones are barely used afterward. `e.title` is never read again, and `e.location` is only used as a truthiness check, which works the same on `safeLocation`. Carrying two representations of the same data is what makes this hard to follow. Keep just the safe fields.

templates/events.html
```diff
@@ -48,6 +49,49 @@ <h1 class="page-title">Events</h1>
...
+        return escapeHtml(url.href);
+      }
+    } catch (error) {
+      if (String(value).charAt(0) === '/') return escapeHtml(value);
+    }
+    return '#';
+  }
+
+  events = events.map(function(e) {
+    var type = Object.prototype.hasOwnProperty.call(eventTypeLabels, e.type) ? e.type : 'third-party';
```
madolson: Nit: the keys are fixed strings from our own frontmatter, so the prototype-pollution guard is defensive noise.

```suggestion
    var type = eventTypeLabels[e.type] ? e.type : 'third-party';
```

templates/events.html
```diff
@@ -48,6 +49,49 @@ <h1 class="page-title">Events</h1>
...
+    try {
+      var url = new URL(value, window.location.origin);
+      if (url.protocol === 'http:' || url.protocol === 'https:') {
+        return escapeHtml(url.href);
+      }
+    } catch (error) {
+      if (String(value).charAt(0) === '/') return escapeHtml(value);
+    }
+    return '#';
+  }
```
madolson, line 67: Minor: `new URL(value, origin)` already resolves relative paths like `/foo`, so the `'/'`-prefix fallback inside the catch is dead code. Simpler:

```suggestion
  function safeHref(value) {
    if (!value) return '';
    try {
      var url = new URL(value, window.location.origin);
      if (url.protocol === 'http:' || url.protocol === 'https:') return escapeHtml(url.href);
    } catch (error) {}
    return '#';
  }
```

madolson (commented):
A few readability suggestions on the template JS, all minor. The event content itself looks good.

*This was generated by AI but verified, with love, by a human.*

madolson (approved):
The current state of the Javascript seems OK to me, so I'll take ownership of any followup if Kyle still has concerns, but would like to see it merged.

PR #594, opened 2026-07-01
Add blog post: Valkey 9.1 Memory Efficiency

### LINKED ISSUE #543: [BLOG POST] memory efficiency
**New Blog Post**

This is an issue created to propose a blog post. Make sure to fill out all the fields so the team can best plan to sequence, edit, and publish your post in a timely manner.
Please replace everything in `[]` before submitting. 

**What is your proposed topic?**
9.1 momentum blog 

**Who is writing this blog post?**
Rain?

**What is your ideal publishing date?**
June 16

**Is this blog post dependent on something else?**
9.1 release and author approval

Andriciuc:
### Description

Adds the technical blog post Valkey 9.1 Memory Efficiency.
 
### Issues Resolved

n/a

### Check List


content/blog/2026-06-24-9.1-memory-efficiency/index.md
```diff
@@ -0,0 +1,100 @@
...
+date = 2026-06-24
+description = "With Valkey 9.1, per-key memory overhead is quietly cut by up to 20% for strings and 7 bytes per sorted set member, without changing a single command or configuration."
+authors = ["dragosandriciuc"]
+[taxonomies]
+blog_type = ["Technical Deep Dive"]
+[extra]
+featured = true
++++
+
+## Reducing Memory Overhead in Valkey 9.1
```
stockholmux: Double heading. Remove.

Image

content/blog/2026-06-24-9.1-memory-efficiency/index.md
```diff
@@ -0,0 +1,100 @@
...
+
+Valkey 9.1 introduces two behind the scenes optimizations that reduce the per-key memory overhead without requiring any new commands or configuration changes.
+
+The two conceptual changes delivered across three PR's below happen deep within Valkey's C data structures, reducing the memory required to store strings and sorted sets, by up to 20% overhead for string keys and 7 bytes for every sorted set member. These changes compound quickly when millions of keys are in use.
+
+This blog explores these changes, and what they mean for your real-world deployments.
+
+## The Cost of Internal Recordkeeping
+
+In Valkey, the values live inside `robj`, a C structure Redis Object. This structure represents not only your data, but also metadata, flags, reference counts and a pointer to the actual value. In the case of `embstr` encoding, the optimized memory encoding for small strings, the string data is immediately allocated after the `robj` header under the form of a contiguous block of memory instead of having two separate memory allocations (for metadata and string value). The simplified layout before Valkey 9.1:
```
stockholmux: ' a C structure Redis Object' -> Let's come up with a better description. We don't tend to use 'Redis' in our content as it's a different project.

stockholmux: '(for metadata and string value)' -> I don't think this needs to be parenthetical

content/blog/2026-06-24-9.1-memory-efficiency/index.md
```diff
@@ -0,0 +1,100 @@
...
+
+![embstr memory before](embstr_memory_before_9.1.svg)
+
+The `ptr` field pointed to the string data that followed the header, however the pointer's value was always deterministic due to the `embstr` encoding guaranteeing that the string is always adjacent to the header, storing this data was redundant.
```
stockholmux: I would better tie these two clauses together 'always adjacent to the header, storing this data was redundant' -> 'always adjacent to the header, **so** storing this data was redundant' or something similar.

content/blog/2026-06-24-9.1-memory-efficiency/index.md
```diff
@@ -0,0 +1,100 @@
...
+
+The two conceptual changes delivered across three PR's below happen deep within Valkey's C data structures, reducing the memory required to store strings and sorted sets, by up to 20% overhead for string keys and 7 bytes for every sorted set member. These changes compound quickly when millions of keys are in use.
+
+This blog explores these changes, and what they mean for your real-world deployments.
+
+## The Cost of Internal Recordkeeping
+
+In Valkey, the values live inside `robj`, a C structure Redis Object. This structure represents not only your data, but also metadata, flags, reference counts and a pointer to the actual value. In the case of `embstr` encoding, the optimized memory encoding for small strings, the string data is immediately allocated after the `robj` header under the form of a contiguous block of memory instead of having two separate memory allocations (for metadata and string value). The simplified layout before Valkey 9.1:
+
+![embstr memory before](embstr_memory_before_9.1.svg)
```
stockholmux, line 33: These look huge. Can you go horizontal instead of vertical?

content/blog/2026-06-24-9.1-memory-efficiency/index.md
```diff
@@ -0,0 +1,100 @@
...
+
+### Eliminating the Redundant Pointer
+
+After the changes in Valkey 9.1, the `ptr` field is removed from the embedded string layout by reusing its 8-byte slot to store the beginning of the embedded string itself. Because the string's location is deterministic, Valkey now computes its address on demand using pointer arithmetic instead of storing an explicit pointer. The new layout looks like this:
+
+![embstr memory after](embstr_memory_after_9.1.svg)
+
+The practical result is that every value stored with `embstr` encoding now saves 8 bytes of overhead. For a workload using 16-byte keys and 16-byte values, this represents roughly a 20% reduction in per-item memory overhead.
+
+### Raising the Embstr Threshold from 64 to 128 Bytes
```
stockholmux: embstr should probably be in code fences. and not capitalized.

content/blog/2026-06-24-9.1-memory-efficiency/index.md
```diff
@@ -0,0 +1,100 @@
...
+
+Removing the above redundant pointer reduced the size of every embedded string object. Because the robj header itself got smaller, this means the same total allocation budget now accommodates more actual string content.
+
+Raising this threshold to 128 bytes (including key, expiry metadata, and value) is additive and this allows the following general benefits:
+
+- fewer allocations per object
```
stockholmux: Let's make this a complete sentence.

When using a colon with bullets, you need to use comma after each bullet, except the last should have a period.

content/blog/2026-06-24-9.1-memory-efficiency/index.md
```diff
@@ -0,0 +1,100 @@
...
+- lower overhead
+- more workloads benefit automatically
+
+Any key returning `embstr` benefits from both of the above optimizations with no extra configuration required.
+
+### Optimizing Sorted Sets to Reduce Overhead
+
+The final improvement in Valkey 9.1 targets Sorted Sets (ZSETs), where the sets also received a memory reduction for workloads that use the `skiplist` representation.
+
+Once a sorted set grows beyond 128 elements, Valkey stores it internally as a `skiplist`. Before Valkey 9.1, each `skiplist` node contained a pointer to a separately allocated SDS string representing the member:
```
stockholmux: 'Once a sorted set grows beyond 128 elements' -> this is configurable, so we should indicate that somewhere e.g. "In the default configuration, once a sorted set..."

content/blog/2026-06-24-9.1-memory-efficiency/index.md
```diff
@@ -0,0 +1,100 @@
...
+
+```txt
++------------------+-------+------------------+---------+-----+---------+
+| element pointer  | score | backward pointer | level-0 | ... | level-N |
++------------------+-------+------------------+---------+-----+---------+
+        |
+        +-------> element SDS
+```
+
+In Valkey 9.1 however, the SDS representation is embedded directly inside the `skiplist` node instead of being stored a separate pointer:
```
stockholmux: You should define SDS. It's also in your diagrams for the embstr sections, so it should be defined there.

stockholmux: 'being stored a separate pointer' -> 'being stored **as** a separate pointer'

content/blog/2026-06-24-9.1-memory-efficiency/index.md
```diff
@@ -0,0 +1,100 @@
...
+
+```txt
++-------+------------------+---------+-----+---------+--------------+
+| score | backward pointer | level-0 | ... | level-N | embedded SDS |
++-------+------------------+---------+-----+---------+--------------+
+```
+
+For sorted sets with thousands or millions of members, this compounds significantly. A `ZSET` with 100,000 members saves roughly 700KB (arithmetic estimate) of memory from this change alone. This change produces a net saving of 7 bytes per sorted set member by removing the pointer.
```
stockholmux: 'roughly 700KB (arithmetic estimate)' drop the '(arithmetic estimate)' since you already say 'roughly'.

stockholmux: I would also flip the order 'This change...' goes before 'A `ZSET`'

content/blog/2026-06-24-9.1-memory-efficiency/index.md
```diff
@@ -0,0 +1,100 @@
...
+
+- fewer allocations per object
+- lower overhead
+- more workloads benefit automatically
+
+Any key returning `embstr` benefits from both of the above optimizations with no extra configuration required.
+
+### Optimizing Sorted Sets to Reduce Overhead
+
+The final improvement in Valkey 9.1 targets Sorted Sets (ZSETs), where the sets also received a memory reduction for workloads that use the `skiplist` representation.
```
stockholmux: you're inconsistent with ZSET, sometimes it's in backticks, others not. Make sure it's all the same way.

content/blog/2026-06-24-9.1-memory-efficiency/index.md
```diff
@@ -0,0 +1,100 @@
...
+For sorted sets with thousands or millions of members, this compounds significantly. A `ZSET` with 100,000 members saves roughly 700KB (arithmetic estimate) of memory from this change alone. This change produces a net saving of 7 bytes per sorted set member by removing the pointer.
+
+## Quantifying the Impact
+
+The above improvements effectiveness vary by workload, key sizes and data type mixes. However, as a practical reference:
+
+| Change | Affected encoding | Saving per item |
+|--------|--------------------|----------------:|
+| Remove `robj->ptr` | `embstr` strings | 8 bytes |
+| Raise `embstr` threshold to 128 bytes | Strings 65–128 bytes (newly eligible for `embstr`) | 8 bytes (newly applies) |
```
stockholmux: 'Strings 65–128 bytes' ... need to say 'length' somewhere

content/blog/2026-06-24-9.1-memory-efficiency/index.md
```diff
@@ -0,0 +1,100 @@
...
+
+## What This Means for Production
+
+These improvements are transparent and you do not need to change your configuration, commands, or application code. The savings activate automatically on upgrade to Valkey 9.1.
+
+For clusters that are operating near memory limits, these no configuration required reductions may bring you back into having a comfortable headroom. However the impact is most pronounced for workloads dominated by small strings or large sorted sets.
+
+With Valkey 9.1 available already, this is the right time to re-examine the `MEMORY USAGE` command and baselines on representative keys and compare them against your 9.0 measurements. The differences will tell you exactly what your specific workload reclaimed.
+
+## Key Takeaways
```
stockholmux: This whole section seems redundant. Replace it with something that provides a more narrative call to action instead of the 1-line "Valkey 9.1 is available now"

content/blog/2026-06-24-9.1-memory-efficiency/three_memory_layout_changes.html
```diff
@@ -0,0 +1,103 @@
+<!DOCTYPE html>
```
stockholmux: I don't understand this file...

stockholmux (changes requested):
Good work. I think it needs some tweaks, but it shouldn't be too bad.

stockholmux:
@Andriciuc you also need to sign off your commits.

content/blog/2026-06-24-9.1-memory-efficiency/index.md
rainsupreme: I agree - "roughly" covers it.

content/blog/2026-06-24-9.1-memory-efficiency/index.md
rainsupreme: Let's call it a ref-counted object, or a serverObject

content/blog/2026-06-24-9.1-memory-efficiency/index.md
```diff
@@ -0,0 +1,100 @@
...
+
+After the changes in Valkey 9.1, the `ptr` field is removed from the embedded string layout by reusing its 8-byte slot to store the beginning of the embedded string itself. Because the string's location is deterministic, Valkey now computes its address on demand using pointer arithmetic instead of storing an explicit pointer. The new layout looks like this:
+
+![embstr memory after](embstr_memory_after_9.1.svg)
+
+The practical result is that every value stored with `embstr` encoding now saves 8 bytes of overhead. For a workload using 16-byte keys and 16-byte values, this represents roughly a 20% reduction in per-item memory overhead.
+
+### Raising the Embstr Threshold from 64 to 128 Bytes
+
+Removing the above redundant pointer reduced the size of every embedded string object. Because the robj header itself got smaller, this means the same total allocation budget now accommodates more actual string content.
```
rainsupreme: nit suggestion: "the above redundant pointer" -> "this redundant pointer"

content/blog/2026-06-24-9.1-memory-efficiency/index.md
```diff
@@ -0,0 +1,100 @@
...
+
+Raising this threshold to 128 bytes (including key, expiry metadata, and value) is additive and this allows the following general benefits:
+
+- fewer allocations per object
+- lower overhead
+- more workloads benefit automatically
+
+Any key returning `embstr` benefits from both of the above optimizations with no extra configuration required.
```
rainsupreme, line 55: returning `embstr` from what command? 🙂 Also, should we brag about how much overhead can be saved here? It's more than 8B often because jemalloc rounds up allocation sizes to certain specific size classes. Reducing the embstr size only helped half of items because some sizes still had to round up to the same size class. Combining allocations can save more than expected because you only have to round up once instead of twice.

OSS Valkey 9.0 vs 9.1: Per-item Overhead Reduction (16B keys)
Directly measured, 5M items per test.

 k+v | 9.0 overhead | 9.1 overhead | Saved | % reduction
-----|--------------|--------------|-------|------------
 24B |       38.9B |       30.9B |  8.0B |        21%
 32B |       38.9B |       30.9B |  8.0B |        21%
 40B |       38.9B |       30.9B |  8.0B |        21%
 48B |       46.9B |       30.9B | 16.0B |        34%
 56B |       46.9B |       38.9B |  8.0B |        17%
 64B |       46.9B |       30.9B | 16.0B |        34%
 72B |       46.9B |       38.9B |  8.0B |        17%
 80B |       54.9B |       30.9B | 24.0B |        44%

Average: 26% overhead reduction
Max: 44% (at k+v=80B, newly-embedded due to threshold adjustment)
Always-embedded sizes (k+v ≤ 40B): 21%

Note: savings vary by exact key+value size due to jemalloc 16B
size-class intervals. ~50% of sizes cross a bucket boundary
(8-24B saved), ~50% absorb the 8B struct shrink within the
same bucket (0B measured change).

content/blog/2026-06-24-9.1-memory-efficiency/index.md
```diff
@@ -0,0 +1,100 @@
...
+|--------|--------------------|----------------:|
+| Remove `robj->ptr` | `embstr` strings | 8 bytes |
+| Raise `embstr` threshold to 128 bytes | Strings 65–128 bytes (newly eligible for `embstr`) | 8 bytes (newly applies) |
+| Embed element SDS inside skiplist node | ZSET members (skiplist encoding) | 7 bytes |
+
+## What This Means for Production
+
+These improvements are transparent and you do not need to change your configuration, commands, or application code. The savings activate automatically on upgrade to Valkey 9.1.
+
+For clusters that are operating near memory limits, these no configuration required reductions may bring you back into having a comfortable headroom. However the impact is most pronounced for workloads dominated by small strings or large sorted sets.
```
rainsupreme, line 97: Also, if you're operating Valkey as a 100% full cache (with an eviction policy like `allkeys-lru`) then your cache suddenly got bigger and your hit rate went up for free. Just saying - up to you if you want to mention this.

content/blog/2026-06-24-9.1-memory-efficiency/three_memory_layout_changes.html
rainsupreme: Glancing over it... is this that struct diagram I had AI put together that I shared? If so, this is terrible and I'm sorry I didn't realize it was creating html instead of svg. This should be replaced. If not and this is something else, I'm also sorry I can't render html in my head 😆

rainsupreme (commented):
It's definitely coming together! 😁

content/blog/2026-06-24-9.1-memory-efficiency/index.md
Andriciuc: Fixed!

content/blog/2026-06-24-9.1-memory-efficiency/index.md
Andriciuc: Fixed, calling it ref-counted object moving forward, thank you!

content/blog/2026-06-24-9.1-memory-efficiency/index.md
Andriciuc: True, sounds much better, added!

content/blog/2026-06-24-9.1-memory-efficiency/index.md
Andriciuc: fixed! also ensured that embstr is code fenced across the blog

content/blog/2026-06-24-9.1-memory-efficiency/index.md
Andriciuc: Thank you Rain! I will use more active voice moving forward. Fixed!

content/blog/2026-06-24-9.1-memory-efficiency/index.md
Andriciuc: Oh! Thank you for the style guide review, will use it as such moving forward, also made it into a sentence.

content/blog/2026-06-24-9.1-memory-efficiency/index.md
Andriciuc: Working on it to see how I can make the svg's smaller, thank you!

content/blog/2026-06-24-9.1-memory-efficiency/index.md
Andriciuc: Fixed! Thank you!

content/blog/2026-06-24-9.1-memory-efficiency/index.md
Andriciuc: Ooooh, I thought wrongly that it was non-configurable, updated!

content/blog/2026-06-24-9.1-memory-efficiency/index.md
Andriciuc: Hmm, I added a definition for the dynamic string below in a note, just to give it more context/info rather than squishing it into the image, and added the `as`. Thank you!

content/blog/2026-06-24-9.1-memory-efficiency/index.md
Andriciuc: 100%! Thank you for adding this, thinking of a way to describe it now.

content/blog/2026-06-24-9.1-memory-efficiency/index.md
Andriciuc: Oh thank you, correct, added Strings with lengths of ...

content/blog/2026-06-24-9.1-memory-efficiency/three_memory_layout_changes.html
Andriciuc: Oh my, I will remake this and probably will remove it, thank you!

content/blog/2026-06-24-9.1-memory-efficiency/index.md
```diff
@@ -0,0 +1,101 @@
...
+
+After the changes in Valkey 9.1, the `ptr` field is removed from the embedded string layout by reusing its 8-byte slot to store the beginning of the embedded string itself. Because the string's location is deterministic, Valkey now computes its address on demand using pointer arithmetic instead of storing an explicit pointer. The new layout looks like this:
+
+![embstr memory after](embstr_memory_after_9.1.svg)
+
+The practical result is that every value stored with `embstr` encoding now saves 8 bytes of overhead. For a workload using 16-byte keys and 16-byte values, this represents roughly a 20% reduction in per-item memory overhead.
+
+### Raising the `embstr` Threshold from 64 to 128 Bytes
+
+Removing this redundant pointer reduced the size of every embedded string object. Because the robj header itself got smaller, this means the same total allocation budget now accommodates more actual string content.
```
stockholmux: `robj` is not styled.

content/blog/2026-06-24-9.1-memory-efficiency/index.md
```diff
@@ -0,0 +1,101 @@
...
+
+Raising this threshold to 128 bytes (including key, expiry metadata, and value) is additive and this allows several benefits such as fewer allocations per object, lower overhead and more workloads benefit automatically.
+
+Any key returning `embstr` benefits from both of the above optimizations with no extra configuration required.
+
+### Optimizing Sorted Sets to Reduce Overhead
+
+The final improvement in Valkey 9.1 targets Sorted Sets (`ZSET`), where the sets also received a memory reduction for workloads that use the `skiplist` representation.
```
stockholmux: I think technically it should be the '`skiplist` encoding'

content/blog/2026-06-24-9.1-memory-efficiency/index.md
```diff
@@ -0,0 +1,101 @@
...
+
+```txt
++------------------+-------+------------------+---------+-----+---------+
+| element pointer  | score | backward pointer | level-0 | ... | level-N |
++------------------+-------+------------------+---------+-----+---------+
+        |
+        +-------> element SDS
+```
+
+!!! note
```
stockholmux: This didn't style properly:

Image

content/blog/2026-06-24-9.1-memory-efficiency/index.md
```diff
@@ -0,0 +1,101 @@
...
+  <img src="embstr_memory_before_9.1.svg" width="500">
+</p>
+
+The `ptr` field pointed to the string data that followed the header, however the pointer's value was always deterministic due to the `embstr` encoding guaranteeing that the string is always adjacent to the header, so storing this data was redundant.
+
+### Eliminating the Redundant Pointer
+
+After the changes in Valkey 9.1, the `ptr` field is removed from the embedded string layout by reusing its 8-byte slot to store the beginning of the embedded string itself. Because the string's location is deterministic, Valkey now computes its address on demand using pointer arithmetic instead of storing an explicit pointer. The new layout looks like this:
+
+![embstr memory after](embstr_memory_after_9.1.svg)
```
stockholmux, line 41: I think we need to make this horizontal (example for layout purposes only... the rest of the styling is 👍 )

Image

stockholmux: On my screen this diagram is a full screen vertically - if we move to horizontal I think it'll present more nicely:

Image

stockholmux (changes requested):
A few more minor issues.

content/blog/2026-06-24-9.1-memory-efficiency/index.md
Andriciuc: Fixed!

content/blog/2026-06-24-9.1-memory-efficiency/index.md
Andriciuc: Done, thank you!

content/blog/2026-06-24-9.1-memory-efficiency/index.md
Andriciuc: Added the encoding mention once. Thank you!

content/blog/2026-06-24-9.1-memory-efficiency/index.md
Andriciuc: Fixed! Thank you, used the _Note: ... _ system for now.

content/blog/2026-06-24-9.1-memory-efficiency/index.md
Andriciuc: I think this is great, thank you for the detailed description! I added this, please let me know if the details are enough.

content/blog/2026-06-24-9.1-memory-efficiency/index.md
Andriciuc: Fixed both images using the new tool, should be much smaller/better to view, thank you!

content/blog/2026-06-24-9.1-memory-efficiency/three_memory_layout_changes.html
Andriciuc: Removed this file.

content/blog/2026-06-24-9.1-memory-efficiency/index.md
Andriciuc: Done!

content/blog/2026-06-24-9.1-memory-efficiency/index.md
Andriciuc: Updated! Please let me know if this is the correct tone, thank you!

content/blog/2026-06-24-9.1-memory-efficiency/index.md
Andriciuc: Done!

stockholmux (approved):
Very pleased with the progress.

content/blog/2026-06-24-9.1-memory-efficiency/index.md
```diff
@@ -0,0 +1,120 @@
...
+For sorted sets with thousands or millions of members, this compounds significantly. A `ZSET` with 100,000 members saves roughly 700KB of memory from this change alone. This change produces a net saving of 7 bytes per sorted set member by removing the pointer.
+
+## Quantifying the Impact
+
+The above improvements effectiveness vary by workload, key sizes and data type mixes. However, as a practical reference:
+
+| Change | Affected encoding | Saving per item |
+|--------|--------------------|----------------:|
+| Remove `robj->ptr` | `embstr` strings | 8 bytes |
+| Raise `embstr` threshold to 128 bytes | Strings with lengths of 65–128 bytes (newly eligible for `embstr`) | 8 bytes (newly applies) |
```
rainsupreme: > Strings with lengths of 65–128 bytes (newly eligible for `embstr`) 

This isn't strictly true, to be perfectly pedantic. The threshold is for the whole robj size including key string, expiry, and other overhead stored in robj, so the actual value string size is somewhat smaller.

rainsupreme (approved):
This is great! I did have one bit of technical pedantry, true, but I like where this is at. :)

content/blog/2026-06-24-9.1-memory-efficiency/embstr_memory_after_9.1.svg
rainsupreme, line 1: nit? the dashed line boarder on this one has transparent fill, while the other has a white fill. The website background is white so it probably doesn't matter... right? 🤷‍♀️

Andriciuc: Done! Update the image and tried to align with the before_9.1.svg🙏

content/blog/2026-06-24-9.1-memory-efficiency/index.md
Andriciuc: Thank you Rain! Fixed by temoving the mention of byte length and generalized the text

content/blog/2026-06-24-9.1-memory-efficiency/index.md
```diff
@@ -0,0 +1,120 @@
...
+| Embed element SDS inside skiplist node | `ZSET` members (skiplist encoding) | 7 bytes |
+
+## What This Means for Production
+
+These improvements are transparent and you do not need to change your configuration, commands, or application code. The savings activate automatically on upgrade to Valkey 9.1.
+
+For clusters that are operating near memory limits, these no configuration required reductions may bring you back into having a comfortable headroom. However the impact is most pronounced for workloads dominated by small strings or large sorted sets.
+
+If you are utilizing Valkey as a dedicated cache running near full capacity with an eviction policy such as `allkeys-lru`, these memory savings effectively increase your cache capacity. With more data fitting into the same amount of memory, your cache will achieve a higher hit rate without requiring any changes.
```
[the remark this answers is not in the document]
Andriciuc: Fixed.

content/blog/2026-06-24-9.1-memory-efficiency/index.md
```diff
@@ -0,0 +1,120 @@
...
+None of the above matters until you check it against real keys. Here's how to find out what Valkey 9.1 is worth to you specifically.
+
+Pick a handful of representative keys from your production deployment and run `MEMORY USAGE` on a 9.0 instance versus a 9.1 instance:
+
+```bash
+> MEMORY USAGE session:abc123
+> OBJECT ENCODING session:abc123
+```
+
+If a string key that used to report `raw` now reports `embstr`, that key just got cheaper to store. No code change, no config change, just a straight upgrade. Multiply the byte difference by your key count and you have the real number, not just an estimate.
```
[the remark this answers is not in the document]
Andriciuc: We direct users to use MEMORY USAGE above. Saving is guaranteed and the "may" here is not necessary.

content/blog/2026-06-24-9.1-memory-efficiency/index.md
```diff
@@ -0,0 +1,120 @@
...
+blog_type = ["Technical Deep Dive"]
+[extra]
+featured = true
++++
+
+Valkey 9.1 introduces two behind the scenes optimizations that reduce the per-key memory overhead without requiring any new commands or configuration changes.
+
+The two conceptual changes delivered across three PR's below happen deep within Valkey's C data structures, reducing the memory required to store strings and sorted sets, by up to 20% overhead for string keys and 7 bytes for every sorted set member. These changes compound quickly when millions of keys are in use.
+
+This blog explores these changes, and what they mean for your real-world deployments.
```
stockholmux: 'blog' -> 'blog post'

stockholmux (approved):
One nitpick but otherwise LGTM.

@crystalphamLF As far as publish date should we go for this week or next? The _Secret Life of Data_ is tomorrow (7/21) and it actually is a good lead to this one. I'd advocate for  7/28 unless you have any objections.

content/blog/2026-06-24-9.1-memory-efficiency/index.md
```diff
@@ -0,0 +1,120 @@
...
+
+## The Cost of Internal Recordkeeping
+
+In Valkey, the values live inside `robj`, a ref-counted object. This structure represents not only your data, but also metadata, flags, reference counts and a pointer to the actual value. In the case of `embstr` encoding, the optimized memory encoding for small strings, the string data is immediately allocated after the `robj` header under the form of a contiguous block of memory instead of having two separate memory allocations for metadata and string value. The simplified layout before Valkey 9.1:
+
+![embstr memory before](embstr_memory_before_9.1.svg)
```
madolson, line 33: This image really doesn't highlight the structure of the memory block, the next one.

content/blog/2026-06-24-9.1-memory-efficiency/index.md
```diff
@@ -0,0 +1,120 @@
++++
+title = "Reducing Memory Overhead in Valkey 9.1"
+date = 2026-06-24
+description = "With Valkey 9.1, per-key memory overhead is quietly cut by up to 20% for strings and 7 bytes per sorted set member, without changing a single command or configuration."
```
madolson: Where did 20% come from? You say 20% for strings and 20% overhead later. We should be consistent.

content/blog/2026-06-24-9.1-memory-efficiency/index.md
```diff
@@ -0,0 +1,120 @@
...
+authors = ["dragosandriciuc"]
+[taxonomies]
+blog_type = ["Technical Deep Dive"]
+[extra]
+featured = true
++++
+
+Valkey 9.1 introduces two behind the scenes optimizations that reduce the per-key memory overhead without requiring any new commands or configuration changes.
```
madolson:
```suggestion
Valkey 9.1 introduces two optimizations that reduce the per-key memory overhead without requiring any new commands or configuration changes.
```
Overall the language is very wordy and flowery. I made some suggestions to make it more concise, which is more aligned with the technical forward tone we want for the blog.

content/blog/2026-06-24-9.1-memory-efficiency/index.md
```diff
@@ -0,0 +1,120 @@
...
+
+The two conceptual changes delivered across three PR's below happen deep within Valkey's C data structures, reducing the memory required to store strings and sorted sets, by up to 20% overhead for string keys and 7 bytes for every sorted set member. These changes compound quickly when millions of keys are in use.
```
madolson:
```suggestion
The three changes reduce the memory required to store strings and sorted sets, by up to 20% overhead for string keys and 7 bytes for every sorted set member. These savings add up quickly when millions of keys are in use.
```
More flower. Compound implies growth, they just add up when there are lots of them.

content/blog/2026-06-24-9.1-memory-efficiency/index.md
madolson:
```suggestion
This blog explores how these optimizations work and what they mean for your real-world deployments.
```

content/blog/2026-06-24-9.1-memory-efficiency/index.md
```diff
@@ -0,0 +1,120 @@
...
+
+This blog explores these changes, and what they mean for your real-world deployments.
+
+## The Cost of Internal Recordkeeping
+
+In Valkey, the values live inside `robj`, a ref-counted object. This structure represents not only your data, but also metadata, flags, reference counts and a pointer to the actual value. In the case of `embstr` encoding, the optimized memory encoding for small strings, the string data is immediately allocated after the `robj` header under the form of a contiguous block of memory instead of having two separate memory allocations for metadata and string value. The simplified layout before Valkey 9.1:
```
madolson:
```suggestion
In Valkey, the values live inside an `robj`, a ref-counted object. This structure represents not only your data, but also metadata, flags, reference counts and a pointer to the actual value. In the case of `embstr` encoding, the optimized memory encoding for small strings, the string data is immediately allocated after the `robj` header under the form of a contiguous block of memory instead of having two separate memory allocations for metadata and string value. The simplified layout before Valkey 9.1:
```

"In the case of `embstr`", encoding is used before encodings are introduced. A connecting sentence explaining what encodings are would be good. I also think the images are not doing a lot of justice to explain what is going on. @rainsupreme you had some fantastic diagrams with memory layouts, can you share/help regenerate those?

content/blog/2026-06-24-9.1-memory-efficiency/index.md
```diff
@@ -0,0 +1,120 @@
...
+
+![embstr memory before](embstr_memory_before_9.1.svg)
+
+The `ptr` field pointed to the string data that followed the header, however the pointer's value was always deterministic due to the `embstr` encoding guaranteeing that the string is always adjacent to the header, so storing this data was redundant.
```
madolson: The obvious question that this raises is, "Why not remove it", because we haven't introduced raw pointers yet. This is never really explained. Raw pointers are sort of hinted at later.

content/blog/2026-06-24-9.1-memory-efficiency/index.md
```diff
@@ -0,0 +1,120 @@
...
+
+### Eliminating the Redundant Pointer
+
+After the changes in Valkey 9.1, the `ptr` field is removed from the embedded string layout by reusing its 8-byte slot to store the beginning of the embedded string itself. Because the string's location is deterministic, Valkey now computes its address on demand using pointer arithmetic instead of storing an explicit pointer. The new layout looks like this:
+
+![embstr memory after](embstr_memory_after_9.1.svg)
+
+The practical result is that every value stored with `embstr` encoding now saves 8 bytes of overhead. For a workload using 16-byte keys and 16-byte values, this represents roughly a 20% reduction in per-item memory overhead.
```
madolson:
```suggestion
The result is that every value stored with `embstr` encoding now saves 8 bytes of overhead. For a workload using 16-byte keys and 16-byte values, this represents roughly a 20% reduction in per-item memory overhead.
```

content/blog/2026-06-24-9.1-memory-efficiency/index.md
```diff
@@ -0,0 +1,120 @@
...
+None of the above matters until you check it against real keys. Here's how to find out what Valkey 9.1 is worth to you specifically.
+
+Pick a handful of representative keys from your production deployment and run `MEMORY USAGE` on a 9.0 instance versus a 9.1 instance:
+
+```bash
+> MEMORY USAGE session:abc123
+> OBJECT ENCODING session:abc123
+```
+
+If a string key that used to report `raw` now reports `embstr`, that key just got cheaper to store. No code change, no config change, just a straight upgrade. Multiply the byte difference by your key count and you have the real number, not just an estimate.
```
madolson:
```suggestion
If a string key that used to report `raw` now reports `embstr`, that key just got cheaper to store. Multiply the byte difference by your key count and you have the real number, not just an estimate.
```

content/blog/2026-06-24-9.1-memory-efficiency/index.md
```diff
@@ -0,0 +1,120 @@
...
+If you are utilizing Valkey as a dedicated cache running near full capacity with an eviction policy such as `allkeys-lru`, these memory savings effectively increase your cache capacity. With more data fitting into the same amount of memory, your cache may achieve a higher hit rate without requiring any changes.
+
+## Try This Against Your Own Workload
+
+None of the above matters until you check it against real keys. Here's how to find out what Valkey 9.1 is worth to you specifically.
+
+Pick a handful of representative keys from your production deployment and run `MEMORY USAGE` on a 9.0 instance versus a 9.1 instance:
+
+```bash
+> MEMORY USAGE session:abc123
```
madolson: You say use `MEMORY USAGE`, then use the output from `ENCODING`.

content/blog/2026-06-24-9.1-memory-efficiency/index.md
```diff
@@ -0,0 +1,120 @@
...
+
+Removing this redundant pointer reduced the size of every embedded string object. Because the `robj` header itself got smaller, this means the same total allocation budget now accommodates more actual string content.
+
+Raising this threshold to 128 bytes (including key, expiry metadata, and value) means more string sizes now qualify for the more compact `embstr` layout instead of falling through the `raw` encoding.
+
+You can check which encoding a given key is using with `OBJECT ENCODING`:
+
+```bash
+> OBJECT ENCODING key
+"embstr"
```
madolson: You explain how to check encoding twice in this blog, should we just do it at the end?

content/blog/2026-06-24-9.1-memory-efficiency/index.md
madolson:
```suggestion
description = "With Valkey 9.1, per-key memory overhead is cut by up to 20% for strings and 7 bytes per sorted set member, without changing a single command or configuration."
```

content/blog/2026-06-24-9.1-memory-efficiency/index.md
```diff
@@ -0,0 +1,120 @@
...
+
+The saving per key is not a flat number. Because Valkey allocates memory through jemalloc, requests are rounded up to fixed size classes rather than granting the exact byte count requested. In practice this produces overhead reductions ranging from roughly 17% to 44%, averaging around 26%.
+
+ _Note: Measured across a range of key+value sizes (16-byte keys), 5 million items per test, Valkey 9.0 compared to 9.1._
+
+The exact saving depends on where your key+value sizes fall relative to jemalloc's size class boundaries.
+
+### Optimizing Sorted Sets to Reduce Overhead
+
+The final improvement in Valkey 9.1 targets Sorted Sets (`ZSET`), where the sets also received a memory reduction for workloads that use the `skiplist` encoding representation.
```
madolson:
```suggestion
The final improvement in Valkey 9.1 targets Sorted Sets, where the sets also received a memory reduction for workloads that use the `skiplist` encoding representation.
```
I don't think we call them ZSET? Zset is a command.

content/blog/2026-06-24-9.1-memory-efficiency/index.md
```diff
@@ -0,0 +1,120 @@
...
++------------------+-------+------------------+---------+-----+---------+
+| element pointer  | score | backward pointer | level-0 | ... | level-N |
++------------------+-------+------------------+---------+-----+---------+
+        |
+        +-------> element SDS
+```
+
+_Note: SDS (Simple Dynamic String) is Valkey's internal  dynamic string implementation. Most keys, string values, and many internal data structures are stored as SDS objects._
+
+In Valkey 9.1 however, the SDS representation is embedded directly inside the `skiplist` node instead of being stored as a separate pointer:
```
madolson: I feel like it's worth explaining why this is a reasonable choice. Keys are typically small, they are immutable, and it's tucked away at the end of the allocation so it doesn't affect the layout of other objects.

madolson (commented):
I like the blog, but would like to see it use a little bit less flowery language. There is also some inconsistency in how things are described, so it would be nice to just be concise.

content/blog/2026-06-24-9.1-memory-efficiency/index.md
cnuthalapati: Standardize to relative -> 20% gives a good idea of the improvement but the 7 is an absolute number & not immediately clear on impact to readers not familiar with Valkey internals

content/blog/2026-06-24-9.1-memory-efficiency/index.md
cnuthalapati, line 1: would be good to quantify impact -> upto xx GB or xx% maybe?

content/blog/2026-06-24-9.1-memory-efficiency/index.md
cnuthalapati: unsure what conditions this happens

content/blog/2026-06-24-9.1-memory-efficiency/index.md
Andriciuc: Thank you! Updated the text, and will keep this in mind moving forward.

content/blog/2026-06-24-9.1-memory-efficiency/index.md
Andriciuc: Updated, thank you for streamlining the paragraph!

content/blog/2026-06-24-9.1-memory-efficiency/index.md
Andriciuc: Fixed!

Andriciuc: Updated.

content/blog/2026-06-24-9.1-memory-efficiency/index.md
Andriciuc: Updated.

content/blog/2026-06-24-9.1-memory-efficiency/index.md
Andriciuc: Updated.

Andriciuc: Hi @cnuthalapati, thank you for the feedback! Could you point to which line this is on? I want to make sure I fix the right conditional, my guess is that it's about when a raw key actually converts to embstr?

content/blog/2026-06-24-9.1-memory-efficiency/index.md
Andriciuc: Ah correct, we do not specify what encoding actually does, so I have updated and added that clarification before the user runs MEMORY USAGE and OBJECT ENCODING. Thank you!

content/blog/2026-06-24-9.1-memory-efficiency/index.md
Andriciuc: Thank you for pointing this out! I updated to "Sorted Sets" throughout for consistency with how Valkey's own docs refer to the data type since zsets is the internal name.

content/blog/2026-06-24-9.1-memory-efficiency/index.md
cnuthalapati: Yes - when does a raw key actually converts to embstr?

content/blog/2026-06-24-9.1-memory-efficiency/index.md
Andriciuc: Thank you! We'll try to provide a better/clearer one soon.

content/blog/2026-06-24-9.1-memory-efficiency/index.md
```diff
@@ -0,0 +1,120 @@
...
+
+![embstr memory after](embstr_memory_after_9.1.svg)
+
+The result is that every value stored with `embstr` encoding now saves 8 bytes of overhead. For a workload using 16-byte keys and 16-byte values, this represents roughly a 20% reduction in per-item memory overhead.
+
+### Raising the `embstr` Threshold from 64 to 128 Bytes
+
+Removing this redundant pointer reduced the size of every embedded string object. Because the `robj` header itself got smaller, this means the same total allocation budget now accommodates more actual string content.
+
+Raising this threshold to 128 bytes (including key, expiry metadata, and value) means more string sizes now qualify for the more compact `embstr` layout instead of falling through the `raw` encoding.
```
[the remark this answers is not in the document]
Andriciuc: Updated to reflect proper OBJECT ENCODING description using value/string length.

content/blog/2026-06-24-9.1-memory-efficiency/index.md
Andriciuc: Thank you! You are correct, given that we already table data that a proper confirmed methodology, I decided to replace the old 20% to a 26% (17–44%, averaging 26% from the actual 9.0 vs 9.1 measured table below)

The original 20% figure came from a single specific case in [PR 2516](https://github.com/valkey-io/valkey/pull/2516) which we mention as references at the end of the blog post.

Andriciuc: Thank you for the feedback, we will look into providing a proper denominator so we can present the information correctly, not only the delta.

content/blog/2026-06-24-9.1-memory-efficiency/index.md
Andriciuc: Hmm, @rainsupreme thoughts on this? We already mention a version of this: "a sorted set with 100,000 members saves roughly 700KB of memory".

content/blog/2026-06-24-9.1-memory-efficiency/index.md
Andriciuc: I have added a paragraph explaining what encodings are for users who are not familiar with the term for Valkey.

content/blog/2026-06-24-9.1-memory-efficiency/index.md
Andriciuc: I have updated the blog to mention OBJECT ENCODING when we first introduce it in the beginning of the blog. Later we explain how in conjunction with memory usage the user can try to see the delta between 9.0 and 9.1 and I'd rather keep that example as is.

content/blog/2026-06-24-9.1-memory-efficiency/index.md
Andriciuc: Hmm, thank you. I have added a short explanatory paragraph of the move. 🙏

content/blog/2026-06-24-9.1-memory-efficiency/index.md
```diff
@@ -0,0 +1,126 @@
...
+
+ _Note: Measured across a range of key+value sizes (16-byte keys), 5 million items per test, Valkey 9.0 compared to 9.1._
+
+The exact saving depends on where your key+value sizes fall relative to jemalloc's size class boundaries.
+
+### Optimizing Sorted Sets to Reduce Overhead
+
+The final improvement in Valkey 9.1 targets Sorted Sets, where the sets also received a memory reduction for workloads that use the `skiplist` encoding representation.
+
+In the default configuration, once a sorted set grows beyond 128 elements, Valkey stores it internally as a `skiplist` encoding. Before Valkey 9.1, each `skiplist` node contained a pointer to a separately allocated SDS string representing the member:
```
rainsupreme: to be technically correct, it also switches to skiplist if a single element is too large (64+ bytes by default)

content/blog/2026-06-24-9.1-memory-efficiency/index.md
```diff
@@ -0,0 +1,126 @@
...
++------------------+-------+------------------+---------+-----+---------+
+| element pointer  | score | backward pointer | level-0 | ... | level-N |
++------------------+-------+------------------+---------+-----+---------+
+        |
+        +-------> element SDS
+```
+
+_Note: SDS (Simple Dynamic String) is Valkey's internal  dynamic string implementation. Most keys, string values, and many internal data structures are stored as SDS objects._
+
+Sorted set members are typically short and immutable. Once a member is added to a sorted set, its value never changes in place, so there is no risk of the embedded string needing to grow and disrupt a nodes layout later. Because the embedded data now sits at the end of the allocation, it does not shift the position of the other fields (score, backward pointer) that the rest of the skiplist logic depends upon.
```
rainsupreme:
```suggestion
Sorted set members are typically short and immutable. Once a member is added to a sorted set, its value never changes in place, so there is no risk of the embedded string needing to grow and disrupt a node's layout later. Because the embedded data now sits at the end of the allocation, it does not shift the position of the other fields (score, backward pointer) that the rest of the skiplist logic depends upon.
```

content/blog/2026-06-24-9.1-memory-efficiency/index.md
Andriciuc: Updated! Improved the sentence contrasting raw encoding against embstr where the pointer is redundant. Thank you for that feedback!

content/blog/2026-06-24-9.1-memory-efficiency/index.md
Andriciuc: Done, the images have been improved, thank you @rainsupreme.

content/blog/2026-06-24-9.1-memory-efficiency/index.md
Andriciuc: Thank you @cnuthalapati for the feedback! I addressed this gap by mentioning that raw converts to embstr once valkey is restarted because RDB and AOF reloads re-apply encoding logic.

content/blog/2026-06-24-9.1-memory-efficiency/index.md
```diff
@@ -32,28 +32,34 @@ One such encoding is `embstr`, the optimized memory encoding for small strings.
...
+
+Raising this threshold to 128 bytes (covering the full robj size including key, expiry metadata, and value) means more string sizes now qualify for the more compact `embstr` layout instead of falling through the `raw` encoding.
+
+![raise embstr threshold after](raise_embstr_threshold_after.svg)
 
 Any key returning `embstr` benefits from both of the above optimizations with no extra configuration required.
 
 The saving per key is not a flat number. Because Valkey allocates memory through jemalloc, requests are rounded up to fixed size classes rather than granting the exact byte count requested. In practice this produces overhead reductions ranging from roughly 17% to 44%, averaging around 26%.
 
  _Note: Measured across a range of key+value sizes (16-byte keys), 5 million items per test, Valkey 9.0 compared to 9.1._
```
rainsupreme: nits:
- We could just remove `(16-byte keys)` as I don't think that detail is important. What matters is key+value size, as you say.
- To make it clear this isn't talking about the chart below we could also make it a parenthetical note as part of the paragraph above.

content/blog/2026-06-24-9.1-memory-efficiency/index.md
rainsupreme: So the cop-out answer is "it depends on element length", due to jemalloc size classes, variable sds header sizes, jemalloc overhead, etc. Here's a bunch of experimental data though!

We measured this directly rather than estimating it. Setup: Valkey 9.0.5 vs 9.1.1, one sorted set populated with 5M members (random insertion order, random scores), `used_memory` delta per member after allowing background reclamation to settle, across 20 member lengths from 10 to 128 bytes. Results were byte-identical across four platforms (Graviton 3/4, AMD, Intel), as expected for allocator-level behavior.

**Summary: for typical short members (10–40 bytes), 9.1 saves ~6–8.5 bytes per member, which is about 11–15% of the per-member overhead, or 7–11% of total per-member memory.**

Example at 20-byte members: 84.8 → 78.3 bytes/member total, of which 28 bytes is user data (member + score), so overhead drops 56.8 → 50.3 bytes (−11.4%).

The saving is not a flat 7 bytes in practice because jemalloc rounds allocations to size classes. Merging the previously-separate member string into the skiplist node means one larger allocation instead of two smaller ones, and larger allocations round in coarser class steps. Depending on where a given member length lands relative to a class boundary, the realized saving ranges from ~11.7 bytes down to zero, and a few specific lengths (44, 60, 96 bytes) actually use slightly more memory in 9.1.

<details>
<summary>Full measured table (bytes per member, 5M members)</summary>

| Member size | 9.0.5 total | 9.1.1 total | Saved | % of overhead | % of total |
|---:|---:|---:|---:|---:|---:|
| 10 | 76.8 | 68.3 | 8.50 | 14.5% | 11.1% |
| 12 | 76.8 | 68.3 | 8.49 | 15.0% | 11.1% |
| 16 | 84.8 | 78.3 | 6.49 | 10.7% | 7.7% |
| 20 | 84.8 | 78.3 | 6.49 | 11.4% | 7.7% |
| 24 | 92.8 | 84.3 | 8.50 | 14.0% | 9.2% |
| 28 | 92.8 | 84.3 | 8.49 | 15.0% | 9.1% |
| 32 | 100.8 | 100.3 | 0.47 | 0.8% | 0.5% |
| 36 | 100.8 | 100.3 | 0.46 | 0.8% | 0.5% |
| 40 | 108.8 | 100.3 | 8.46 | 13.9% | 7.8% |
| 44 | 108.8 | 116.5 | −7.71 | −13.6% | −7.1% |
| 48 | 116.8 | 116.5 | 0.31 | 0.5% | 0.3% |
| 52 | 116.8 | 116.5 | 0.31 | 0.6% | 0.3% |
| 56 | 124.8 | 116.4 | 8.31 | 13.7% | 6.7% |
| 60 | 124.8 | 133.1 | −8.28 | −14.6% | −6.6% |
| 64 | 140.8 | 133.1 | 7.70 | 11.2% | 5.5% |
| 72 | 140.8 | 133.1 | 7.70 | 12.7% | 5.5% |
| 80 | 156.8 | 151.5 | 5.29 | 7.7% | 3.4% |
| 96 | 172.8 | 177.1 | −4.29 | −6.2% | −2.5% |
| 112 | 188.8 | 183.5 | 5.30 | 7.7% | 2.8% |
| 128 | 220.8 | 209.1 | 11.69 | 13.8% | 5.3% |

Overhead = total minus user data (member length + 8-byte score). Averaged uniformly over all 20 sizes the saving is ~4.1 bytes/member; weighted toward the short members that dominate real workloads it is meaningfully higher.
</details>

For the blog title/description, a phrasing that stays honest to the data:

> "saves up to 8.5 bytes per sorted set member — around 11–15% of per-member overhead for typical short members"

with the body noting that the exact saving depends on member length relative to allocator size classes (same caveat the post already makes for strings).

To be clear, a lot of the detail here is so we (author + reviewers) understand clearly - I think we only really need the punchline in the article.

content/blog/2026-06-24-9.1-memory-efficiency/index.md
```diff
@@ -62,37 +68,27 @@ The final improvement in Valkey 9.1 targets Sorted Sets, where the sets also rec
...
-| Change | Affected encoding | Saving per item |
-|--------|--------------------|----------------:|
-| Remove `robj->ptr` | `embstr` strings | 8 bytes |
-| Raise `embstr` threshold to 128 bytes | Objects newly eligible under the raised threshold | 8 bytes (newly applies) |
-| Embed element SDS inside skiplist node | Sorted set members (skiplist encoding) | 7 bytes |
+| Change | Affected encoding | Struct-level change | Measured overhead reduction |
+|--------|--------------------|---:|---:|
+| Remove `robj->ptr` | `embstr` strings | -8 bytes | 17–44% (avg. 26%) |
+| Raise `embstr` threshold to 128 bytes | Objects newly eligible under the raised threshold | -8 bytes (newly applies) | included in range above |
+| Embed element SDS inside skiplist node | Sorted set members (skiplist encoding) | -7 bytes | not separately measured |
```
rainsupreme: I think the data I just brought does separately measure the skiplist change now, so you can update this line :)
https://github.com/valkey-io/valkey-io.github.io/pull/594/changes/5f7163e4c390aea600300dbed3b3510a2016679c..ddf8f4d954d4d2d15fd51747fb6896d66bf5ea43#r3640769938

Btw I do like the column you added here

content/blog/2026-06-24-9.1-memory-efficiency/index.md
Andriciuc: Updated, thank you!

content/blog/2026-06-24-9.1-memory-efficiency/index.md
Andriciuc: Updated, thank you.

content/blog/2026-06-24-9.1-memory-efficiency/index.md
Andriciuc: Thank you! Added parentheses and removed the 16 byte mention.

content/blog/2026-06-24-9.1-memory-efficiency/index.md
Andriciuc: Thank you Rain! Updated with proper data, I always like displaying such changes in graphs as it seems easier to view rather than lines of text.

content/blog/2026-06-24-9.1-memory-efficiency/index.md
Andriciuc: Thank you for confirming the statistics @rainsupreme! @cnuthalapati I have updated the whole blog to reflect the above findings, mentioning short members and the proper 8.5 bytes per sorted set member across the blog post.

content/blog/2026-06-24-9.1-memory-efficiency/index.md
Andriciuc: Fixed as per Rain's recent statistics

content/blog/2026-06-24-9.1-memory-efficiency/index.md
```diff
@@ -0,0 +1,122 @@
...
+None of the above matters until you check it against real keys. Here's how to find out what Valkey 9.1 is worth to you specifically.
+
+Pick a handful of representative keys from your production deployment and run `MEMORY USAGE` and `OBJECT ENCODING` on a 9.0 instance versus a 9.1 instance where `MEMORY USAGE` gives you the byte count, `OBJECT ENCODING` tells you which layout produced it:
+
+```bash
+> MEMORY USAGE session:abc123
+> OBJECT ENCODING session:abc123
+```
+
+A key converts from `raw` to `embstr` once its combined key+expiry+value size falls within the new 128-byte budget. A restart alone is enough to trigger it, since RDB and AOF reload re-applies encoding logic on 9.1. If a key reports `embstr` where it used to report `raw`, that key just got cheaper to store. Multiply the byte difference by your key count and you have the real number, not just an estimate.
```
stockholmux: 'since RDB and AOF reload re-applies encoding logic on 9.1' 

This is true but it implies that it doesn't re-apply logic on previous versions. So, drop the 'on 9.1' and it will be more accurate.

stockholmux (changes requested):
One small nit then LGTM.

content/blog/2026-06-24-9.1-memory-efficiency/index.md
Andriciuc: Oh yes, updated!

content/blog/2026-06-24-9.1-memory-efficiency/index.md
```diff
@@ -0,0 +1,122 @@
...
+
+### Raising the `embstr` Threshold from 64 to 128 Bytes
+
+Removing this redundant pointer reduced the size of every embedded string object. Because the `robj` header itself got smaller, this means the same total allocation budget now accommodates more actual string content.
+
+![raise embstr threshold before](raise_embstr_threshold_before.svg)
+
+Raising this threshold to 128 bytes (covering the full robj size including key, expiry metadata, and value) means more string sizes now qualify for the more compact `embstr` layout instead of falling through the `raw` encoding.
+
+![raise embstr threshold after](raise_embstr_threshold_after.svg)
```
madolson, line 53: I don't think this math would make it embedded?

```
8    sizeof(robj) - sizeof(void *)
1    prefixed sds hdr size byte
18   key: 16B + 1B sdshdr5 + NUL
130  value: 126B + 3B sdshdr8 + NUL
---
157
```

content/blog/2026-06-24-9.1-memory-efficiency/index.md
```diff
@@ -0,0 +1,122 @@
...
+date = 2026-07-28
+description = "With Valkey 9.1, per-key memory overhead is cut by up to 44% (averaging 26%) for strings and up to 8.5 bytes per sorted set member, without changing a single command or configuration."
+authors = ["dragosandriciuc"]
+[taxonomies]
+blog_type = ["Technical Deep Dive"]
+[extra]
+featured = true
++++
+
+Valkey 9.1 introduces two optimizations that reduce the per-key memory overhead without requiring any new commands or configuration changes.
```
madolson: Reads as an error. "two optimizations" here, then three subsections, three table rows, three reference PRs, and the sorted set work is "The final improvement," which implies a third.

Just say three, or "two optimizations across three changes."

content/blog/2026-06-24-9.1-memory-efficiency/index.md
```diff
@@ -0,0 +1,122 @@
...
+
+For sorted sets, the savings depend on your member lengths where short members (10 to 40 bytes) see the most benefit, roughly 11-15% of per-member overhead. Compare `MEMORY USAGE` on a representative sorted set before and after upgrading to see what your actual member sizes yield, since the exact number varies by length in the same allocator-rounding way string savings do.
+
+The memory you get back by simply moving to 9.1 and measuring what has changed is the real upgrade.
+
+## Reference PRs
+
+- [PR 2516 — Remove redundant robj->ptr for embstr encoding](https://github.com/valkey-io/valkey/pull/2516)
+- [PR 3397 — Raise embstr threshold from 64 to 128 bytes](https://github.com/valkey-io/valkey/pull/3397)
+- [PR 2508 — Embed ele in skiplist node for `ZSET` members](https://github.com/valkey-io/valkey/pull/2508)
```
madolson: Nit, re-raising. Still `ZSET` here. The body is fixed.

content/blog/2026-06-24-9.1-memory-efficiency/index.md
```diff
@@ -0,0 +1,122 @@
...
+None of the above matters until you check it against real keys. Here's how to find out what Valkey 9.1 is worth to you specifically.
+
+Pick a handful of representative keys from your production deployment and run `MEMORY USAGE` and `OBJECT ENCODING` on a 9.0 instance versus a 9.1 instance where `MEMORY USAGE` gives you the byte count, `OBJECT ENCODING` tells you which layout produced it:
+
+```bash
+> MEMORY USAGE session:abc123
+> OBJECT ENCODING session:abc123
+```
+
+A key converts from `raw` to `embstr` once its combined key+expiry+value size falls within the new 128-byte budget. A restart alone is enough to trigger it, since RDB and AOF reload re-applies encoding logic. If a key reports `embstr` where it used to report `raw`, that key just got cheaper to store. Multiply the byte difference by your key count and you have the real number, not just an estimate.
```
madolson: Optional. Dropping the 700KB example killed the only absolute number in the post, which is what @cnuthalapati was after.

@rainsupreme's table has a measured one: 5M members at 20 bytes goes 84.8 to 78.3 bytes/member, so roughly 32MB back on one sorted set. Measured, not extrapolated.

content/blog/2026-06-24-9.1-memory-efficiency/index.md
```diff
@@ -0,0 +1,122 @@
++++
+title = "Reducing Memory Overhead in Valkey 9.1"
+date = 2026-07-28
+description = "With Valkey 9.1, per-key memory overhead is cut by up to 44% (averaging 26%) for strings and up to 8.5 bytes per sorted set member, without changing a single command or configuration."
```
madolson:
```suggestion
description = "With Valkey 9.1, per-key memory overhead is cut by up to 44% for strings and up to 8.5 bytes per sorted set member, without changing a single command or configuration."
```

Saying "averaging 26%" implies the average user will see 26%, which is hard to justify. The up to 44% is the hedge I'm OK with, since it's the best case.

content/blog/2026-06-24-9.1-memory-efficiency/index.md
```diff
@@ -0,0 +1,122 @@
...
+authors = ["dragosandriciuc"]
+[taxonomies]
+blog_type = ["Technical Deep Dive"]
+[extra]
+featured = true
++++
+
+Valkey 9.1 introduces two optimizations that reduce the per-key memory overhead without requiring any new commands or configuration changes.
+
+These reduce the memory required to store strings and sorted sets, averaging 26% overhead for string keys and up to 8.5 bytes per sorted set member for typical short members. These savings add up quickly when millions of keys are in use.
```
madolson:
```suggestion
These reduce the memory required to store strings and sorted sets, up to 44% overhead for string keys and up to 8.5 bytes per sorted set member for typical short members. These savings add up quickly when millions of keys are in use.
```

madolson (approved):
I'm going to approve since I'm broadly happy, but I do think we should fix the 26% comment. Also some assorted minor comments, but I think the diagrams make the flow a lot better!

crystalphamLF:
@valkey-io/technical-blog-reviewers may we push? It is not letting me merge on my end

stockholmux (approved):
LGTM

content/blog/2026-06-24-9.1-memory-efficiency/index.md
Andriciuc: Fixed! Thank you.

content/blog/2026-06-24-9.1-memory-efficiency/index.md
Andriciuc: removed ZSET and kept the title intact as per the PR. Thank you!

content/blog/2026-06-24-9.1-memory-efficiency/index.md
Andriciuc: @rainsupreme is it possible to provide an updated image for the above results? Thank you!

crystalphamLF:
Thank you all! This one has been published. Sharing on social shortly

PR #599, opened 2026-07-14
Add RFC 9116 security.txt to static/.well-known/

kobihikri:
## What this does

Adds `/.well-known/security.txt` to the site by placing it in `static/.well-known/security.txt`, which Hugo copies verbatim to the published site root.

## Why

[RFC 9116](https://www.rfc-editor.org/rfc/rfc9116) defines a small, standard file at `/.well-known/security.txt` so a security researcher can find where to report a vulnerability without guessing. Right now the live file is missing:

```
$ curl -sSL -o /dev/null -w '%{http_code}\n' https://valkey.io/.well-known/security.txt
404
```

The project already has a security contact — I just pointed the file at the existing mailbox and policy rather than inventing anything:

- `Contact: mailto:security@lists.valkey.io` (from `SECURITY.md` in this repo)
- `Policy: https://github.com/valkey-io/valkey/security/policy`
- `Expires:` — RFC 9116 requires this field; I set it ~1 year out.

## Verifying it will serve

`static/` is already served verbatim — for example `static/img/IconGithub.svg` is live at `https://valkey.io/img/IconGithub.svg` (200). Hugo copies dot-directories under `static/` too, so `static/.well-known/security.txt` will publish at `https://valkey.io/.well-known/security.txt`. Please double-check on a preview build.

Single file, no other changes. Commit is DCO signed-off.

---

*Disclosure: I used an AI tool to help spot this and draft the file. I verified every line myself against the live site and this repo's `SECURITY.md`, and I take responsibility for the change. Happy to adjust the `Expires` date, contact, or policy link to whatever the maintainers prefer.*

### DIFF
--- a/.github/workflows/zola-deploy.yml
+++ b/.github/workflows/zola-deploy.yml
@@ -54,7 +54,7 @@ jobs:
           repository: valkey-io/valkey-json
           path: valkey-json
 
-      - name: Init commands, topics and clients
+      - name: Init commands, topics, clients and security.txt
         run: |
           cd website
           ./build/init-topics-and-clients.sh ../valkey-doc/topics \
@@ -62,6 +62,7 @@ jobs:
           ./build/init-commands.sh ../valkey-doc/commands \
             ../valkey/src/commands ../valkey-bloom/src/commands \
             ../valkey-json/src/commands ../valkey-search/src/commands
+          ./build/init-security-txt.sh .
 
       - name: Build only
         uses: shalzz/zola-deploy-action@v0.22.0
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ _data/modules.json
 .idea/*
 tmp/*
 static/debug
+static/.well-known/security.txt
--- a/build/init-security-txt.sh
+++ b/build/init-security-txt.sh
@@ -1,0 +1,42 @@
+#!/bin/bash
+# See README for usage
+# This file will generate static/.well-known/security.txt (RFC 9116).
+#
+# The Expires field is derived from the build date rather than committed, so it
+# always sits one year past the last site build and cannot silently go stale.
+
+# first check to make sure there are arguments
+if [ -z "$1" ]; then
+    echo "You must supply a path to the site root as the first argument"
+    exit 1
+fi
+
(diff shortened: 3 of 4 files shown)


stockholmux:
1. This site isn't a Hugo site.
2. The static expiry seems like a silly thing to just forget to update. Use a template to make sure it always stays 1 year + last site build.

kobihikri:
Both points taken, @stockholmux — thanks, and you were right on the first one.

**1. Not Hugo.** My PR description said Hugo twice; this is a **Zola** site (`config.toml` with `compile_sass`/`build_search_index`/`taxonomies`, `templates/` and `sass/` directories, `zola-deploy.yml`). I checked rather than assumed this time. The `static/` passthrough behaviour I relied on happens to be the same in both, so the file would still have served — but I stated the wrong reason, and that's a fair thing to call out.

**2. The static expiry.** Agreed, and fixed in 8ba6721 — it's now derived from the build instead of committed:

- `build/init-security-txt.sh` generates `static/.well-known/security.txt` with `Expires` set one year past the build, following the style of the existing scripts in `build/`.
- It's invoked from the same workflow step that already runs `init-topics-and-clients.sh` and `init-commands.sh`, so it rides along with every build with no new step.
- The generated file is gitignored, matching how `content/commands/*` and `content/topics/*` are handled — so there's no committed date to forget about.

One note on mechanism, since you said "use a template": I looked at doing this as a Zola template first, but Zola only emits templated output at fixed paths it knows about (`index`, `sitemap`, `robots.txt`, feeds) — a page with `path = "/.well-known/security.txt"` produces `\.well-known/security.txt/index.html`, not the file itself. Generating before `zola build` was the closest way to get "1 year + last site build" at the exact path. If you'd rather it lived somewhere else, or want the window to be something other than a year, that's a one-line change.

Also worth flagging honestly: the script branches on GNU vs BSD `date`, and I could only execute the BSD branch locally. The GNU branch is what CI will actually run, so this PR's build is the real check on it.

Disclosure: I used AI assistance (Claude Opus 5) in preparing this. I verified the generator against this repo's conventions and ran it myself, and I take responsibility for it.

PR #611, opened 2026-07-24
Add blog post: Valkey 9.1 benchmark

Andriciuc:
### Description

This PR is a blog post that describes valkey's benchmarking tool. It provides a how to, going through 3 levers that the user can use to benchmark valkey on their system.
 
### Check List
- [ X] Commits are signed per the DCO using `--signoff`

### DIFF
--- a/content/blog/2026-08-15-what-is-valkey-benchmark/index.md
+++ b/content/blog/2026-08-15-what-is-valkey-benchmark/index.md
@@ -9,15 +9,15 @@ blog_type = ["How-to"]
 featured = true
 +++
 
-Evaluating performances for database instances is a critical part in understanding the potential weight your infrastructure can carry. And Valkey leverages this with an out of the box easy to use tool that allows you to get actionable figures for your instance and to evaluate that performance.
+Benchmarking your database instances is one of the simplest ways to understand how your infrastructure behaves under load and where its limits are. And Valkey leverages this with an easy-to-use tool that allows you to get actionable figures for your instance and to evaluate that performance.
 
 This blog post explores the functions of the tool and some common traps you might encounter on your first run.
 
-## What is valkey-benchmark?
+## What is `valkey-benchmark`?
 
-Valkey ships with valkey-benchmark, a simple CLI load-generation tool that simulates multiple clients sending commands at a Valkey instance on your hardware. The tool helps you answer "how does my server perform under load", and not "how fast is Valkey" in your environment.
+Valkey ships with `valkey-benchmark`, a simple CLI load-generation tool that simulates multiple clients sending commands at a Valkey instance on your hardware. The tool helps you answer "how does my server perform under load", and not "how fast is Valkey" in your environment.
 
-You can use the tool through a simple invocation, such as valkey-benchmark with no args against localhost through multiple connections to the server, giving you a quick benchmark against your system.
+You can use the tool through a simple invocation, such as `valkey-benchmark` with no arguments against localhost through multiple connections to the server, giving you a quick benchmark against your system.
 
 For example, you can run the benchmark with the default configuration against 127.0.0.1:6379:
 
@@ -36,20 +36,110 @@ And the example result is:
         
 _NOTE: You need to have a running Valkey instance before launching the benchmark._
 
-Its default behavior does not stress the server much, however, it does more than just a simple basic invocation and benchmark run.
+That number looks impressive, but it doesn't tell the whole story of what the benchmarking tool is capable of.
 
 ### The default run isn't your workload
 
-Rarely is a simple invocation as the one above a good benchmark for a system under load. The 67k req/s hit the same single key. There is no real key-space pressure, no cache-miss behavior and no memory access patterns resembling real life production environments.
+A bare invocation like this rarely reflects how your system behaves under real load. The 67k req/s hit the same single key. There is no real key-space pressure, no cache-miss behavior and no memory access patterns resembling real life production environments.
 
 The actual payload was 3 bytes, which is certainly smaller than anything most applications actually store. With no pipelining, one in-flight request per connection, it resembles more a network-bound worst case rather than anything a real pipelining client sees.
 
-However, the tool's number is not wrong. It answers a more focused question that most clients assume, "how fast can 50 simple, unpipelined connections hit one key?", not "how will my app perform". Which has it merit as a standard benchmark procedure.
+However, the tool's number is not wrong. It answers a more focused question that most people assume, "how fast can 50 simple, unpipelined connections hit one key?", not "how will my app perform".
 
-And none of these default invocations are bad, they're just not realistic to your workload. Here's how you can make valkey-benchmark change that.
+And the default invocation is not bad, they're just not representative of your workload. Here's how you can make `valkey-benchmark` do that.
 
-### Three levers that make it realistic
+## Three ways to make your benchmark resemble production
 
-### Cluster mode
+It's important to point out the golden rule of a useful benchmark before moving forward, a useful benchmark compares apples to apples. In this case, you can compare different versions of Valkey on the same workload or the same version of Valkey, but with different options.
+
+### Simulate a real keyspace
+
+By default, `valkey-benchmark` repeatedly accesses the same key. Most applications operate over thousands or millions of keys. A realistic keyspace introduces cache misses, different memory access patterns, and a workload that more closely resembles production.
+
+You can generate these realistic keyspace benchmarks using random keys with the `-r` option:
+
+        ```bash
+        valkey-benchmark -t set -r 100000 -n 1000000
+        ```
+
+Instead of repeatedly updating a single key, each operation selects a random key from a space of 100,000 possible keys. After one million requests you end up with roughly 100,000 keys stored, closely resembling an application repeatedly writing into an existing dataset.
+
+The important point isn't the exact number of keys. It is choosing a keyspace that resembles your own application. A session cache, leaderboard, and job queue all exercise memory differently.
+
+In the above case, the example result is:
+
+        ```bash
+        Summary:
+                throughput summary: 103734.44 requests per second
+                latency summary (msec):
+                        avg       min       p50       p95       p99       max
+                        0.279     0.072     0.231     0.567     1.007     3.615
+        ```
+
+_NOTE: The exact numbers vary from run to run, the point is the change in what's being tested, not a numeric ranking._
+
+### Use realistic payload transmissions
+
+The default `valkey-benchmark` stores values that are 3 bytes long. It keeps the benchmark lightweight, but realistically session data, cached API responses and serialized objects often reach bigger numbers of bytes.
+
+However, the payload size often does not matter until you cross a certain threshold, for example the ethernet packet size (~1500 bytes).
+
+These larger values not only affect memory consumption, but also increase network bandwidth requiring more memory to copy and allocate, which expose more performance characteristics than the default payloads.
+
+You can change the generated value size with the `-d` option. Combining this with a realistic keyspace (-r) allows you to simulate a workload that is representative of a realistic application rather than the benchmark's default configuration:
+
+        ```bash
+        valkey-benchmark -t set -r 100000 -n 1000000 -d 1024
+        ```
+
+The result numbers vary, but a 1024-byte payload often has only a modest impact on throughput because it remains below the Ethernet MTU in some environments.
+
+        ```bash
+        Summary:
+                throughput summary: 111135.80 requests per second
+                latency summary (msec):
+                        avg       min       p50       p95       p99       max
+                        0.263     0.088     0.215     0.559     1.175     5.119
+        ```
+
+### Pipeline requests like a real client
+
+Most production clients do not wait for every command to finish before sending the next one. Instead, they batch several requests together using pipelining, which reduces the cost of network round trips.
+
+In Valkey, the default benchmark has pipelining disabled, meaning each connection has only one outstanding request. However, modern applications do not have this type of conservative, network-bound workload.
+
+Adding even a modest pipeline dramatically changes the picture, and combined with a realistic keyspace and payload size, pipelining provides a much stronger baseline for evaluating your infrastructure:
+
+        ```bash
+        valkey-benchmark -t set -r 100000 -n 1000000 -d 1024 -P 16
+        ```
+
+This benchmark measures how efficiently your system processes batches of commands through Valkey. Depending on your workload, throughput can increase several times over simply by matching the pipelining strategy used by your application.
+
+        ```bash
+        Summary:
+                throughput summary: 448631.66 requests per second
+                latency summary (msec):
+                        avg       min       p50       p95       p99       max
+                        1.632     0.224     1.311     3.455     5.943    13.991
+        ```
+
+As with every benchmark parameter, realism is more valuable than chasing the highest requests-per-second number. Choose a pipeline depth that reflects what your client library actually uses.
+
+Together, these three settings, keyspace size (-r), pipeline depth (-P), and payload size (-d) move `valkey-benchmark` away from a simple benchmark and toward a workload that more closely resembles production.
+
+## Cluster mode
+
+If you want to benchmark your application with data that is automatically sharded across multiple Valkey cluster nodes, cluster mode is the solution.
+
+All of the above benchmarks can be done by adding the `--cluster` option. When running cluster mode with a custom command, your key names need to include `{tag}`, a placeholder that ensures the command routes to the correct node. Combined with `--rfr`, this lets you stress-test read-heavy workloads specifically.
 
 ## Putting it all together
+
+Between a realistic keyspace, representative payloads, and pipelining that matches your client, `valkey-benchmark` can move from a synthetic microbenchmark to something that actually predicts how your application behaves under load. The tool's own goal is reproducibility, so anchor your numbers to a fixed set of options, and compare against your own past runs rather than other benchmark tools.
+
+You can copy and paste this following benchmark command as your starting point:
+
+        ```bash
+        valkey-benchmark -t set,get -r 100000 -d 1024 -P 16 -n 1000000 -q
+        ```
--- a/content/blog/2026-08-15-what-is-valkey-benchmark/index.md
+++ b/content/blog/2026-08-15-what-is-valkey-benchmark/index.md
@@ -1,120 +1,55 @@
 +++
-title = "Reducing Memory Overhead in Valkey 9.1"
-date = 2026-06-24
-description = "With Valkey 9.1, per-key memory overhead is quietly cut by up to 20% for strings and 7 bytes per sorted set member, without changing a single command or configuration."
+title = "Simulating Real Workloads with valkey-benchmark"
+date = 2026-07-31
+description = "A practical guide to shaping valkey-benchmark's default run into a realistic workload simulation"
 authors = ["dragosandriciuc"]
 [taxonomies]
-blog_type = ["Technical Deep Dive"]
+blog_type = ["How-to"]
 [extra]
 featured = true
 +++
 
-Valkey 9.1 introduces two behind the scenes optimizations that reduce the per-key memory overhead without requiring any new commands or configuration changes.
+Evaluating performances for database instances is a critical part in understanding the potential weight your infrastructure can carry. And Valkey leverages this with an out of the box easy to use tool that allows you to get actionable figures for your instance and to evaluate that performance.
 
-The two conceptual changes delivered across three PR's below happen deep within Valkey's C data structures, reducing the memory required to store strings and sorted sets, by up to 20% overhead for string keys and 7 bytes for every sorted set member. These changes compound quickly when millions of keys are in use.
+This blog post explores the functions of the tool and some common traps you might encounter on your first run.
 
-This blog explores these changes, and what they mean for your real-world deployments.
+## What is valkey-benchmark?
 
-## The Cost of Internal Recordkeeping
+Valkey ships with valkey-benchmark, a simple CLI load-generation tool that simulates multiple clients sending commands at a Valkey instance on your hardware. The tool helps you answer "how does my server perform under load", and not "how fast is Valkey" in your environment.
 
-In Valkey, the values live inside `robj`, a ref-counted object. This structure represents not only your data, but also metadata, flags, reference counts and a pointer to the actual value. In the case of `embstr` encoding, the optimized memory encoding for small strings, the string data is immediately allocated after the `robj` header under the form of a contiguous block of memory instead of having two separate memory allocations for metadata and string value. The simplified layout before Valkey 9.1:
+You can use the tool through a simple invocation, such as valkey-benchmark with no args against localhost through multiple connections to the server, giving you a quick benchmark against your system.
 
-![embstr memory before](embstr_memory_before_9.1.svg)
+For example, you can run the benchmark with the default configuration against 127.0.0.1:6379:
 
-The `ptr` field pointed to the string data that followed the header, however the pointer's value was always deterministic due to the `embstr` encoding guaranteeing that the string is always adjacent to the header, so storing this data was redundant.
+        ```bash
+        $ valkey-benchmark
+        ```
+And the example result is:
 
-### Eliminating the Redundant Pointer
+        ```bash
+        Summary:
+                throughput summary: 67204.30 requests per second
+                latency summary (msec):
+                avg       min       p50       p95       p99       max
+                0.471     0.080     0.343     1.015     1.999    12.007
+        ```
+        
+_NOTE: You need to have a running Valkey instance before launching the benchmark._
 
-After the changes in Valkey 9.1, the `ptr` field is removed from the embedded string layout by reusing its 8-byte slot to store the beginning of the embedded string itself. Because the string's location is deterministic, Valkey now computes its address on demand using pointer arithmetic instead of storing an explicit pointer. The new layout looks like this:
+Its default behavior does not stress the server much, however, it does more than just a simple basic invocation and benchmark run.
 
-![embstr memory after](embstr_memory_after_9.1.svg)
+### The default run isn't your workload
 
-The practical result is that every value stored with `embstr` encoding now saves 8 bytes of overhead. For a workload using 16-byte keys and 16-byte values, this represents roughly a 20% reduction in per-item memory overhead.
+Rarely is a simple invocation as the one above a good benchmark for a system under load. The 67k req/s hit the same single key. There is no real key-space pressure, no cache-miss behavior and no memory access patterns resembling real life production environments.
 
-### Raising the `embstr` Threshold from 64 to 128 Bytes
+The actual payload was 3 bytes, which is certainly smaller than anything most applications actually store. With no pipelining, one in-flight request per connection, it resembles more a network-bound worst case rather than anything a real pipelining client sees.
 
-Removing this redundant pointer reduced the size of every embedded string object. Because the `robj` header itself got smaller, this means the same total allocation budget now accommodates more actual string content.
+However, the tool's number is not wrong. It answers a more focused question that most clients assume, "how fast can 50 simple, unpipelined connections hit one key?", not "how will my app perform". Which has it merit as a standard benchmark procedure.
 
-Raising this threshold to 128 bytes (including key, expiry metadata, and value) means more string sizes now qualify for the more compact `embstr` layout instead of falling through the `raw` encoding.
+And none of these default invocations are bad, they're just not realistic to your workload. Here's how you can make valkey-benchmark change that.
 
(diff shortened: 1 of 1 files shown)


content/blog/2026-08-15-what-is-valkey-benchmark/index.md
```diff
@@ -0,0 +1,146 @@
...
+date = 2026-07-31
+description = "A practical guide to shaping valkey-benchmark's default run into a realistic workload simulation"
+authors = ["dragosandriciuc"]
+[taxonomies]
+blog_type = ["How-to"]
+[extra]
+featured = true
++++
+
+Benchmarking your database instances is one of the simplest ways to understand how your infrastructure behaves under load and where its limits are. And Valkey leverages this with an easy-to-use tool that allows you to get actionable figures for your instance and to evaluate that performance.
```
stockholmux: 'database instances' is a weird wording here. I would just just say "Benchmarking your Valkey deployment'

stockholmux: Also, I wouldn't use 'instance' as a descriptor throughout.

content/blog/2026-08-15-what-is-valkey-benchmark/index.md
```diff
@@ -0,0 +1,146 @@
...
+
+This blog post explores the functions of the tool and some common traps you might encounter on your first run.
```
stockholmux, line 14: 'This blog post explores' is the same phrase you used on your last one. Mix up the language a bit.

content/blog/2026-08-15-what-is-valkey-benchmark/index.md
```diff
@@ -0,0 +1,146 @@
...
+
+## What is `valkey-benchmark`?
+
+Valkey ships with `valkey-benchmark`, a simple CLI load-generation tool that simulates multiple clients sending commands at a Valkey instance on your hardware. The tool helps you answer "how does my server perform under load", and not "how fast is Valkey" in your environment.
```
stockholmux: Remove simple throughout (I don't like easy/simple etc. - that's in the eye of the beholder)

content/blog/2026-08-15-what-is-valkey-benchmark/index.md
```diff
@@ -0,0 +1,146 @@
...
+
+You can use the tool through a simple invocation, such as `valkey-benchmark` with no arguments against localhost through multiple connections to the server, giving you a quick benchmark against your system.
```
stockholmux: I'm not sure I understand what you mean with 'localhost through multiple connections'

content/blog/2026-08-15-what-is-valkey-benchmark/index.md
```diff
@@ -0,0 +1,146 @@
...
+Summary:
+        throughput summary: 67204.30 requests per second
+        latency summary (msec):
+        avg       min       p50       p95       p99       max
+        0.471     0.080     0.343     1.015     1.999    12.007
+```
+
+_NOTE: You need to have a running Valkey instance before launching the benchmark._
+
+That number looks impressive, but it doesn't tell the whole story of what the benchmarking tool is capable of.
```
stockholmux: I don't think it looks impressive! Then later you say ' it resembles more a network-bound worst case rather than anything a real pipelining client sees'

I think I would refactor this into something like "What story does that number tell?" or similar.

content/blog/2026-08-15-what-is-valkey-benchmark/index.md
```diff
@@ -0,0 +1,146 @@
...
+
+### The default run isn't your workload
+
+A bare invocation like this rarely reflects how your system behaves under real load. The 67k req/s hit the same single key. There is no real key-space pressure, no cache-miss behavior and no memory access patterns resembling real life production environments.
+
+The actual payload was 3 bytes, which is certainly smaller than anything most applications actually store. With no pipelining, one in-flight request per connection, it resembles more a network-bound worst case rather than anything a real pipelining client sees.
+
+However, the tool's number is not wrong. It answers a more focused question that most people assume, "how fast can 50 simple, unpipelined connections hit one key?", not "how will my app perform".
+
+And the default invocation is not bad, they're just not representative of your workload. Here's how you can make `valkey-benchmark` do that.
```
stockholmux: 'And the default invocation is not bad, they're just not representative of your workload.' - odd use of 'they're' - I don't think of a default invocation as plural.

'Here's how you can make `valkey-benchmark` do that.' feels out of place. I think you can remove it.

content/blog/2026-08-15-what-is-valkey-benchmark/index.md
```diff
@@ -0,0 +1,146 @@
...
+
+## Three ways to make your benchmark resemble production
+
+It's important to point out the golden rule of a useful benchmark before moving forward, a useful benchmark compares apples to apples. In this case, you can compare different versions of Valkey on the same workload or the same version of Valkey, but with different options.
```
stockholmux: 'It's important to point out ' feels superfluous. It would be strong by just starting out with "The golden rule..."

content/blog/2026-08-15-what-is-valkey-benchmark/index.md
```diff
@@ -0,0 +1,146 @@
...
+
+### Simulate a real keyspace
+
+By default, `valkey-benchmark` repeatedly accesses the same key. Most applications operate over thousands or millions of keys. A realistic keyspace introduces cache misses, different memory access patterns, and a workload that more closely resembles production.
+
+You can generate these realistic keyspace benchmarks using random keys with the `-r` option:
```
stockholmux: I would call it a 'more realistic'

content/blog/2026-08-15-what-is-valkey-benchmark/index.md
```diff
@@ -0,0 +1,146 @@
...
+        latency summary (msec):
+                avg       min       p50       p95       p99       max
+                0.279     0.072     0.231     0.567     1.007     3.615
+```
+
+_NOTE: The exact numbers vary from run to run, the point is the change in what's being tested, not a numeric ranking._
+
+### Use realistic payload transmissions
+
+The default `valkey-benchmark` stores values that are 3 bytes long. It keeps the benchmark lightweight, but realistically session data, cached API responses and serialized objects often reach bigger numbers of bytes.
```
stockholmux: 'often reach bigger numbers of bytes' is a bit awkward. How about 'are often much larger'.

content/blog/2026-08-15-what-is-valkey-benchmark/index.md
```diff
@@ -0,0 +1,146 @@
...
+
+However, the payload size often does not matter until you cross a certain threshold, for example the ethernet packet size (~1500 bytes).
```
stockholmux: source?

content/blog/2026-08-15-what-is-valkey-benchmark/index.md
```diff
@@ -0,0 +1,146 @@
...
+
+For example, you can run the benchmark with the default configuration against 127.0.0.1:6379:
+
+```bash
+$ valkey-benchmark
+```
+
+And the example result is:
+
+```bash
```
stockholmux: use "```text" for these instead of bash. Bash makes it weirdly colourful!

content/blog/2026-08-15-what-is-valkey-benchmark/index.md
```diff
@@ -0,0 +1,146 @@
...
+
+These larger values not only affect memory consumption, but also increase network bandwidth requiring more memory to copy and allocate, which expose more performance characteristics than the default payloads.
+
+You can change the generated value size with the `-d` option. Combining this with a realistic keyspace (-r) allows you to simulate a workload that is representative of a realistic application rather than the benchmark's default configuration:
+
+```bash
+valkey-benchmark -t set -r 100000 -n 1000000 -d 1024
+```
+
+The result numbers vary, but a 1024-byte payload often has only a modest impact on throughput because it remains below the Ethernet MTU in some environments.
```
stockholmux: Unless you're describing which environments, don't call that out, instead just say it _often_ remains.

content/blog/2026-08-15-what-is-valkey-benchmark/index.md
```diff
@@ -0,0 +1,146 @@
...
+Summary:
+        throughput summary: 448631.66 requests per second
+        latency summary (msec):
+                avg       min       p50       p95       p99       max
+                1.632     0.224     1.311     3.455     5.943    13.991
+```
+
+As with every benchmark parameter, realism is more valuable than chasing the highest requests-per-second number. Choose a pipeline depth that reflects what your client library actually uses.
+
+Together, these three settings, keyspace size (-r), pipeline depth (-P), and payload size (-d) move `valkey-benchmark` away from a simple benchmark and toward a workload that more closely resembles production.
```
stockholmux: you've only focused on throughput so far but the examples show latency. That should be addressed somehow.

content/blog/2026-08-15-what-is-valkey-benchmark/index.md
```diff
@@ -0,0 +1,146 @@
...
+
+## Cluster mode
+
+If you want to benchmark your application with data that is automatically sharded across multiple Valkey cluster nodes, cluster mode is the solution.
```
stockholmux: I would rephrase that, 'cluster mode' typically refers to the way you run Valkey. Here I think you're talking about Valkey Benchmark's cluster argument?

content/blog/2026-08-15-what-is-valkey-benchmark/index.md
```diff
@@ -0,0 +1,146 @@
...
+
+All of the above benchmarks can be done by adding the `--cluster` option. When running cluster mode with a custom command, your key names need to include `{tag}`, a placeholder that ensures the command routes to the correct node. Combined with `--rfr`, this lets you stress-test read-heavy workloads specifically.
```
stockholmux: Say more on the `{tag}` thing because I don't get what you're saying.

content/blog/2026-08-15-what-is-valkey-benchmark/index.md
```diff
@@ -0,0 +1,146 @@
...
+
+## Putting it all together
+
+Between a realistic keyspace, representative payloads, and pipelining that matches your client, `valkey-benchmark` can move from a synthetic microbenchmark to something that actually predicts how your application behaves under load. The tool's own goal is reproducibility, so anchor your numbers to a fixed set of options, and compare against your own past runs rather than other benchmark tools.
+
+You can copy and paste this following benchmark command as your starting point:
+
+```bash
+valkey-benchmark -t set,get -r 100000 -d 1024 -P 16 -n 1000000 -q
+```
```
stockholmux, line 163: Needs a CTA

stockholmux (changes requested):
Good start! Two big callouts: 
- Cover latency
- Work on your summary and CTA.

content/blog/2026-08-15-what-is-valkey-benchmark/index.md
Andriciuc: Thank you! I will replace instance with deployment. Updated.

content/blog/2026-08-15-what-is-valkey-benchmark/index.md
Andriciuc: Ah yes, thank you, I will mix it up more!

content/blog/2026-08-15-what-is-valkey-benchmark/index.md
Andriciuc: Exactly yes, thank you! Updated

content/blog/2026-08-15-what-is-valkey-benchmark/index.md
Andriciuc: Done! I will keep this in mind moving forward. 🙏

content/blog/2026-08-15-what-is-valkey-benchmark/index.md
Andriciuc: Thank you! I managed to make it ambiguous, I initially wanted to say something like "...against your local deployment using multiple client connections...", so I'll update it to make it clearer.

Also, I updated invocation mentions with run, I feel it reads better / easier across the blog.

content/blog/2026-08-15-what-is-valkey-benchmark/index.md
Andriciuc: Oh, done! Updated to text throughout.

content/blog/2026-08-15-what-is-valkey-benchmark/index.md
Andriciuc: Updated.

content/blog/2026-08-15-what-is-valkey-benchmark/index.md
Andriciuc: Simplification is always appreciated, updated! Thank you.

content/blog/2026-08-15-what-is-valkey-benchmark/index.md
Andriciuc: Updated.

content/blog/2026-08-15-what-is-valkey-benchmark/index.md
```diff
@@ -0,0 +1,146 @@
...
+## What is `valkey-benchmark`?
+
+Valkey ships with `valkey-benchmark`, a simple CLI load-generation tool that simulates multiple clients sending commands at a Valkey instance on your hardware. The tool helps you answer "how does my server perform under load", and not "how fast is Valkey" in your environment.
+
+You can use the tool through a simple invocation, such as `valkey-benchmark` with no arguments against localhost through multiple connections to the server, giving you a quick benchmark against your system.
+
+For example, you can run the benchmark with the default configuration against 127.0.0.1:6379:
+
+```bash
+$ valkey-benchmark
```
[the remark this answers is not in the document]
Andriciuc: fixed

content/blog/2026-08-15-what-is-valkey-benchmark/index.md
```diff
@@ -0,0 +1,146 @@
...
+
+By default, `valkey-benchmark` repeatedly accesses the same key. Most applications operate over thousands or millions of keys. A realistic keyspace introduces cache misses, different memory access patterns, and a workload that more closely resembles production.
+
+You can generate these realistic keyspace benchmarks using random keys with the `-r` option:
+
+```bash
+valkey-benchmark -t set -r 100000 -n 1000000
+```
+
+Instead of repeatedly updating a single key, each operation selects a random key from a space of 100,000 possible keys. After one million requests you end up with roughly 100,000 keys stored, closely resembling an application repeatedly writing into an existing dataset.
```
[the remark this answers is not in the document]
Andriciuc: Added as a note below the paragraph: 

_NOTE: On an empty database, the above command produces approximately 100,000 stored keys. Existing keys may be overwritten, you can verify this with the `DBSIZE` command._

content/blog/2026-08-15-what-is-valkey-benchmark/index.md
Andriciuc: I see exactly what you mean, in fact... I will stay away from such claims in the future. Saying simple/impressive/etc is subjective, thank you.

I updated this to hook better into the next part, thank you!

content/blog/2026-08-15-what-is-valkey-benchmark/index.md
Andriciuc: Ah, done. Fixed the `they're` and replaced with `it's`. I also replaced "your workload" with "a realistic" workload, as we cannot assume what the client's workload is, I made it more general.

Also removed the mention at the end, thank you!

content/blog/2026-08-15-what-is-valkey-benchmark/index.md
Andriciuc: Updated, understood!

content/blog/2026-08-15-what-is-valkey-benchmark/index.md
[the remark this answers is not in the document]
Andriciuc: I have added the -c arg with 6 and key-{tag} as well as -rfr yes and put a note to let the reader know this works only if they have deployed a valkey cluster

Andriciuc: I reorganized how we present the information to the reader and ensured that -c and {tag} is presented as changes from the standard benchmarking command. Thank you!

content/blog/2026-08-15-what-is-valkey-benchmark/index.md
Andriciuc: I found this in the Valkey benchmarking docs page itself, https://valkey.io/topics/benchmark/, specifically the paragraph starting with "When an ethernet network is used to access Valkey..."

However, I trimmed this paragraph too much, I will update the paragraph to better get the point across, thank you!

content/blog/2026-08-15-what-is-valkey-benchmark/index.md
```diff
@@ -0,0 +1,161 @@
...
+        0.471     0.080     0.343     1.015     1.999    12.007
+```
+
+_NOTE: You need to have a running Valkey deployment before launching the benchmark._
+
+That's a real number, given how simple this default run is, but it's worth asking exactly what it's measuring before you trust it for capacity planning.
+
+### The default run isn't your workload
+
+A bare run like the above result rarely reflects how your system behaves under real load. The 67k req/s hit the same single key. There is no real key-space pressure, no cache-miss behavior and no memory access patterns resembling real life production environments.
```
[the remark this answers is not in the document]
Andriciuc: Done!

content/blog/2026-08-15-what-is-valkey-benchmark/index.md
```diff
@@ -0,0 +1,161 @@
...
+
+The actual payload was 3 bytes, which is certainly smaller than anything most applications actually store. With no pipelining, one in-flight request per connection, it resembles a network-bound worst case rather than anything a real pipelining client would produce.
+
+However, it answers a more focused question that most people assume, "how fast can 50, unpipelined connections hit one key?", not "how will my application perform".
+
+And the default run is not bad, it's just not representative of a realistic workload.
+
+## Three ways to make your benchmark resemble production
+
+The golden rule of a useful benchmark before moving forward, a useful benchmark compares apples to apples. In this case, you can compare different versions of Valkey on the same workload or the same version of Valkey, but with different options.
```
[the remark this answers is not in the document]
Andriciuc: Fixed!

content/blog/2026-08-15-what-is-valkey-benchmark/index.md
```diff
@@ -0,0 +1,161 @@
...
+
+_NOTE: The exact numbers vary from run to run, the point is the change in what's being tested, not a numeric ranking._
+
+### Use realistic payload transmissions
+
+The default `valkey-benchmark` stores values that are 3 bytes long. It keeps the benchmark lightweight, but realistically session data, cached API responses and serialized objects are often much larger.
+
+However, the payload size often does not matter until you cross a certain threshold once you're pipelining. Because 10-byte, 100-byte, and 1000-byte payloads all produce roughly the same throughput under pipelining, with the effect breaking down once payloads approach the ethernet packet size (~1500 bytes).
+
+These larger values not only affect memory consumption, but also increase network bandwidth requiring more memory to copy and allocate, which expose more performance characteristics than the default payloads.
```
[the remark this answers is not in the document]
Andriciuc: fixed!

content/blog/2026-08-15-what-is-valkey-benchmark/index.md
Andriciuc: Thank you for that mention, I reworded the paragraph to include latency mentions between p50 1.311 and p50 0.215 due to pipelining's aggregate throughput

content/blog/2026-08-15-what-is-valkey-benchmark/index.md
```diff
@@ -0,0 +1,161 @@
...
+_NOTE: This example requires an actual Valkey cluster deployment, it does not run against a standalone instance._
+
+## Putting it all together
+
+Between a realistic keyspace, representative payloads, and pipelining that matches your client, `valkey-benchmark` can move from a synthetic microbenchmark to something that actually predicts how your application behaves under load. The tool's own goal is reproducibility, so anchor your numbers to a fixed set of options, and compare against your own past runs rather than other benchmark tools.
+
+You can copy and paste this following benchmark command as your starting point:
+
+```text
+valkey-benchmark -t set,get -r 100000 -d 1024 -P 16 -n 1000000 -q
```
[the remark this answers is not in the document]
Andriciuc: Added a note about -q, describing why it is used.

content/blog/2026-08-15-what-is-valkey-benchmark/index.md
Andriciuc: Updated the ending paragraph encouraging the reader to try the above commands out, and providing them an example to start with updated with a note for the q arg

content/blog/2026-08-15-what-is-valkey-benchmark/index.md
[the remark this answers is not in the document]
Andriciuc: The --seed argument is out of scope for this case, explaining too many arguments in one blog post is not what is required.

content/blog/2026-08-15-what-is-valkey-benchmark/index.md
```diff
@@ -0,0 +1,165 @@
...
+        latency summary (msec):
+                avg       min       p50       p95       p99       max
+                1.632     0.224     1.311     3.455     5.943    13.991
+```
+
+The trade-off is that the throughput roughly quadruples, but per-request latency also rises: p50 climbs from 0.215ms to 1.311ms. That's expected as pipelining trades individual request latency for aggregate throughput, since each request waits in a batch before being sent.
+
+This is exactly why matching your pipeline depth to your actual client matters: a pipeline depth that's unrealistically deep shows great throughput numbers while your real requests wait longer than your application can tolerate.
+
+Together, these three settings: keyspace size (-r), pipeline depth (-P), and payload size (-d), move `valkey-benchmark` away from an ordinary benchmark and toward a workload that more closely resembles production.
```
stockholmux, line 134: maybe 'ordinary' should be 'unrealistic'?

content/blog/2026-08-15-what-is-valkey-benchmark/index.md
stockholmux: I think it needs a more directive CTA.  Something like: 'If you want to find out more...' (link to docs), or 'Share your benchmarking tips in a blog post', etc.

stockholmux (changes requested):
Getting very close

content/blog/2026-08-15-what-is-valkey-benchmark/index.md
Andriciuc: replaced with unrealistic, thank you!

content/blog/2026-08-15-what-is-valkey-benchmark/index.md
Andriciuc: Added a path to take the user to benchmarking documentation and calling out specific commands that are pertinent to this blog if they'd need more information. Thank you! 

I have a solid idea how to use CTA's after these changes. 🙏

content/blog/2026-08-15-what-is-valkey-benchmark/index.md
```diff
@@ -0,0 +1,167 @@
...
+
+- `-c` (client connections) must be at least the number of nodes in your cluster; with fewer clients than nodes, some nodes will never get contacted.
+- Custom commands need `{tag}` in the key name, a placeholder that ensures the command routes to the correct node.
+
+If you also want to stress read replicas, add the `--rfr` (read-from-replicas) argument, with a mode of `no`, `yes`, or `all`. Use this only with read commands as writes sent to a replica are rejected.
+
+For example, against a 6-node cluster (3 primaries and 3 replicas), reading from replicas only:
+
+```text
+valkey-benchmark --cluster -c 6 -n 100000 --rfr yes get key-{tag}
```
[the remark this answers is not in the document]
Andriciuc: Creating a cluster to test configuration now.

stockholmux (approved):
Overall this looks good to me.

crystalphamLF:
@Andriciuc Great! Can you change the date to August 25th? Better posting time for the blogs. :) Thank you! cc: @stockholmux

content/blog/2026-08-15-what-is-valkey-benchmark/index.md
```diff
@@ -0,0 +1,167 @@
...
+
+_NOTE: The exact numbers vary from run to run, the point is the change in what's being tested, not a numeric ranking._
+
+### Use realistic payload transmissions
+
+The default `valkey-benchmark` stores values that are 3 bytes long. It keeps the benchmark lightweight, but realistically session data, cached API responses and serialized objects are often much larger.
+
+However, the payload size often does not matter until you cross a certain threshold once you're pipelining because 10-byte, 100-byte, and 1000-byte payloads all produce roughly the same throughput under pipelining, with the effect breaking down once payloads approach the ethernet packet size (~1500 bytes).
+
+These larger values not only affect memory consumption, but also increase network bandwidth usage requiring more memory to copy and allocate, which expose more performance characteristics than the default payloads.
```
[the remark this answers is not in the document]
Andriciuc: The payload size often has little effect until you cross a certain threshold once you're pipelining, because 10-byte, 100-byte, and 1000-byte payloads all produce roughly the same throughput when accessing Valkey over an ethernet network, with the effect breaking down once payloads approach the ethernet packet size (~1500 bytes).

And for the second mention:

A similar effect applies over ethernet networks specifically: 10-byte, 100-byte, and 1000-byte payloads all produce roughly the same throughput, with the effect breaking down near the ethernet packet size (~1500 bytes).

I will commit these changes shortly, I also updated does not to little effect.

content/blog/2026-08-15-what-is-valkey-benchmark/index.md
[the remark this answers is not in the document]
Andriciuc: Having ran the issue through a 3 master/replica setup I have identified the issue and will push an update for the command now, I will replace the above command with: valkey-benchmark --cluster -h 127.0.0.1 -p 6371 -c 6 -n 100000 --rfr yes get key-{tag}

This should work.

Andriciuc:
> @Andriciuc Great! Can you change the date to August 25th? Better posting time for the blogs. :) Thank you! cc: @stockholmux

@crystalphamLF Done, thank you! Also resolved all issues left above.

content/blog/2026-08-15-what-is-valkey-benchmark/index.md
```diff
@@ -0,0 +1,167 @@
...
+
+```text
+Summary:
+        throughput summary: 448631.66 requests per second
+        latency summary (msec):
+                avg       min       p50       p95       p99       max
+                1.632     0.224     1.311     3.455     5.943    13.991
+```
+
+The trade-off is that the throughput roughly quadruples, but per-request latency also rises: p50 climbs from 0.215ms to 1.311ms. That's expected as pipelining trades individual request latency for aggregate throughput, since each request waits in a batch before being sent.
```
madolson: I don't know if this is a fair categorization. The reason throughput goes up is because the system is more efficient at parsing 4 packets then 1 packet at a time, and the commands are all executed together. The latency is mostly an artifact. You also shouldn't just trust the latency produced by the benchmark, since it works by sending as much traffic as fast as possible. Pipelining causes the traffic to be burstier, which artificially causes queuing which spikes latency. If you use the`--rps` argument, you can should see the latency stay pretty tight around the RTT.

content/blog/2026-08-15-what-is-valkey-benchmark/index.md
```diff
@@ -0,0 +1,167 @@
...
+
+By default, `valkey-benchmark` repeatedly accesses the same key. Most applications operate over thousands or millions of keys. A realistic keyspace introduces cache misses, different memory access patterns, and a workload that more closely resembles production.
+
+You can generate more realistic keyspace benchmarks using random keys with the `-r` option:
+
+```text
+valkey-benchmark -t set -r 100000 -n 1000000
+```
+
+Instead of repeatedly updating a single key, each operation selects a random key from a space of 100,000 possible keys. After one million requests you end up with roughly 100,000 keys stored, closely resembling an application repeatedly writing into an existing dataset.
```
madolson, line 66: The keysize is a little small for benchmarking. It's about 100MB of data, which falls cleanly outside typical L1/L2 cache sizes, but can fit inside some large L3s or at least have a high percentage in L3 cache. The folks that do benchmarking using bump of the key magnitude to ~1,000,000, since that is always cleanly outside of L3 cache hit rates. 

Perhaps can be addressed just as a note.

PR #622, opened 2026-07-31
Auto-publish scheduled blog posts on their date

madolson:
## What

Adds an hourly GitHub Actions workflow that publishes blog posts once their date has arrived, so maintainers can merge a post ahead of time without it going live immediately.

## How

- Contributors/maintainers set `draft = true` in a post's frontmatter and set `date` to the intended publish date/time (UTC).
- `.github/workflows/publish-scheduled-blogs.yml` runs hourly. `build/publish-scheduled-blogs.py` scans `content/blog`, and for any draft whose `date` has passed it removes the `draft` line and commits to `main`, which triggers the existing deploy.

This uses Zola's **native** `draft` support (not a custom field), so a pending post is excluded from the build, the blog listing, the sitemap, and the RSS feed until it is released. `zola serve --drafts` still previews it locally.

## Why the app token

The commit is pushed with the `valkeyrie-bot` app token rather than `GITHUB_TOKEN`. Pushes made with `GITHUB_TOKEN` do not trigger other workflows, which would commit the post without ever rebuilding the site.

## Notes / things to check before merge

- The workflow needs `secrets.VALKEYRIE_BOT_*` usable from a `schedule` trigger and the app to have `contents: write` on this repo. Recommend a first run via **Run workflow** with `dry_run` ticked to confirm.
- `main` is currently unprotected, so the bot can push directly. If branch protection is added later, this will need a bypass or a PR-based flow.
- Drafts with no `date` are never auto-published, so work-in-progress can safely sit on `main`.

## Verified

- Script tested against the real repo's frontmatter variants (`date =` / `date=`, with/without times), nested `index.md` posts, `draft` under a `[extra]` table (correctly ignored), missing dates, and idempotency.
- End-to-end: a draft post was confirmed absent from build/listing/sitemap/feed, then present after the script released it and the site rebuilt.
- `zola serve` with and without `--drafts` confirmed the live vs. preview behavior.

This was generated by AI but verified, with love, by a human.

### DIFF
--- a/CONTRIBUTING-BLOG-POST.md
+++ b/CONTRIBUTING-BLOG-POST.md
@@ -83,6 +83,10 @@ description= "It's become clear that people want to talk about Valkey and have b
 # 'authors' are the folks who wrote or contributed to the post.
 # Each author corresponds to a biography file (more info later in this document)
 authors= [ "maury", "jacobim" ]
+# 'draft' holds the post back from being published.
+# While this is true, the post is not built, listed, or included in the feed.
+# Leave it out unless you are scheduling; see "Scheduling a post" below.
+draft = true
 [extra]
     # 'featured' controls whether the blog post appears in the featured section on the main blog page
     featured = true
@@ -157,6 +161,16 @@ Make sure to communicate your change fully in the body of the pull request.
 After your contribution is made, the website maintainers will review the post.
 They may have feedback for you and ask you to make changes.
 Once everyone is satisfied with the post, the maintainers will merge it into the `main` branch.
-The `main` branch of the repo represents the *future* state of all integrated changes before publishing.
-At this point, the maintainers make further changes to your post to properly schedule or link the post.
-Once this occurs, the maintainers will make move the changes into ‘production’ which will trigger a rebuild and publishing of the content website to [Valkey.io](http://valkey.io/)
+Merging to `main` rebuilds and publishes [Valkey.io](http://valkey.io/), so a post without `draft = true` goes live as soon as it is merged.
+
+## Scheduling a post
+
+To merge a post before it should be public, set `draft = true` in the frontmatter and set `date` to the intended publish date and time (UTC).
+Zola excludes drafts entirely: the post is not built, does not appear in the blog listing, and stays out of the sitemap and RSS feed.
+
+An hourly job (`.github/workflows/publish-scheduled-blogs.yml`) checks the drafts under `content/blog`.
+Once a draft's `date` has passed, the job removes the `draft` line and commits to `main`, which publishes the post.
+Expect the post to go live within an hour of its `date`.
+
+A draft with no `date` is never published automatically, so it is safe to park work-in-progress on `main`.
+Maintainers can also publish immediately by removing the `draft` line by hand, or run the workflow from the Actions tab with **Run workflow** (tick `dry_run` to preview what would be published).
--- a/CONTRIBUTING-BLOG-POST.md
+++ b/CONTRIBUTING-BLOG-POST.md
@@ -71,11 +71,10 @@ Here is an example of the frontmatter:
 +++
 # `title` is how your post will be listed and what will appear at the top of the post
 title=  "Using Valkey for mind control experiments"
-# `date` is when your post will be published.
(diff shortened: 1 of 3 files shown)


build/publish-scheduled-blogs.py
```diff
@@ -0,0 +1,118 @@
...
+Only the `draft` line is touched; the rest of the file is left byte-for-byte
+intact. Writes the list of released files to $GITHUB_OUTPUT as `published`.
+"""
+
+import datetime
+import os
+import pathlib
+import re
+import sys
+import tomllib
```
stockholmux: Okay. Quite confused. you're importing tomllib but you're also using regular expressions to manipulate the TOML?

build/publish-scheduled-blogs.py
```diff
@@ -0,0 +1,118 @@
...
+
+UTC = datetime.timezone.utc
+FRONTMATTER = re.compile(r"^\+\+\+[^\S\n]*\n(.*?)\n\+\+\+[^\S\n]*$", re.S | re.M)
+DRAFT_LINE = re.compile(r"^\s*draft\s*=\s*true\s*(#.*)?$")
```
stockholmux: There should be a better way of extracting this given you're using `tomllib`

build/publish-scheduled-blogs.py
```diff
@@ -0,0 +1,118 @@
...
+            continue
+
+        if data.get("draft") is not True:
+            continue
+
+        if "date" not in data:
+            print(f"::warning file={path}::draft has no date; leaving as draft")
+            continue
+
+        publish_at = as_utc(data["date"], path)
```
stockholmux: This will change the blog process a bit. We'll need to be very explicit that times are in UTC. Right now everything is just set of middle of the night on `date` and we manually push it. If we continue with this we'll get middle-of-the-night posting.

CONTRIBUTING-BLOG-POST.md
```diff
@@ -157,6 +161,16 @@ Make sure to communicate your change fully in the body of the pull request.
...
+## Scheduling a post
+
+To merge a post before it should be public, set `draft = true` in the frontmatter and set `date` to the intended publish date and time (UTC).
+Zola excludes drafts entirely: the post is not built, does not appear in the blog listing, and stays out of the sitemap and RSS feed.
+
+An hourly job (`.github/workflows/publish-scheduled-blogs.yml`) checks the drafts under `content/blog`.
+Once a draft's `date` has passed, the job removes the `draft` line and commits to `main`, which publishes the post.
+Expect the post to go live within an hour of its `date`.
+
+A draft with no `date` is never published automatically, so it is safe to park work-in-progress on `main`.
```
stockholmux: I'd rather not merge work-in-progress. I think we should remove this.

stockholmux (changes requested):
I'm willing to give this a try but I think this is going to be a foot-gun for unintended publishing. 

As for the code itself, I'd rather not see regular expression parsing on toml but rather the proper parsing library you're already using.

Finally, I think we have some process issues: we shouldn't merge work in progress and we need to make sure the blog process is very, very explicit about UTC posting times.

build/publish-scheduled-blogs.py
madolson: I do everything in UTC, is PT maybe better? My original idea was just run the job once a day.

CONTRIBUTING-BLOG-POST.md
madolson: Ok, sure, this came up once related to another PR we handed off, but this was mostly an after thought not a strong requirement.

build/publish-scheduled-blogs.py
madolson: I pivoted half way through, this is a fair callout.

madolson:
@stockholmux Aight. Tried to make less of a foot gun. Always publishes at 15:00 UTC on the defined date, so we can schedule stuff out and not have to worry about individual availability. Rewrote a chunk of TOML parsing to use more of another dependency, to do less stuff manual.

stockholmux (approved):
LGTM

PR #557, opened 2026-05-23
Add valkey-GLIDE PHP client to recommended clients list

adrianluna:
## Description

Adds the Valkey GLIDE PHP client to the recommended clients list on the [clients page](https://valkey.io/clients/#php).

This is the companion change to valkey-io/valkey-doc#443, which adds the `clients/php/valkey-GLIDE.json` metadata file. This PR ensures the website renders the new entry by including it in `recommended_clients_paths`.

## Changes

- Added `/php/valkey-GLIDE.json` to `recommended_clients_paths` in
  `content/clients/_index.md`

## Testing

Screenshot 2026-05-22 at 22 17 07

PR #650, opened 2026-08-28
Add Valkey GLIDE PHP and Ruby clients to recommended clients list

Aryex:
### Description

Adds the Valkey GLIDE PHP and Valkey GLIDE Ruby clients to the recommended clients list on `/clients/`.

**Note:** Supersedes #557, which adds the PHP entry only. Happy to drop the PHP line here instead if @adrianluna would rather land that one.


Aryex:
Hi @madolson @stockholmux, could we can a review for these 🙏

madolson (approved):

PR #656, opened 2026-09-01
Unfurl blog posts with their own image, not the logo

madolson:
Every blog post unfurls with the Valkey logo. `og:image` is hardcoded in `templates/includes/head.html:9` and ignores the `featured_image` the page already renders as its hero. Now uses `featured_image` when set, logo otherwise, and makes the URL absolute since a relative `og:image` is out of spec.

Zola autoescapes the interpolated path, so the tag renders `&#x2F;` instead of `/`. Scrapers decode it, same as the `` on every existing post.

`zola build`: all 63 blog pages point at an image file that exists. Posts with no `featured_image`, sections, and the homepage still get the logo.

*This was generated by AI but verified, with love, by a human.*

### DIFF
--- a/templates/includes/head.html
+++ b/templates/includes/head.html
@@ -6,7 +6,7 @@
     <meta name="MSSmartTagsPreventParsing" content="true" />
     <meta name="Copyright" content="" />
     <meta name="keywords" content="open-source, valkey" />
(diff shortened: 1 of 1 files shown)


stockholmux (approved):
LGTM

PR #659, opened 2026-09-02
Add blog post: Valkey 9.2 new commands

Andriciuc:
### Description

This PR adds a new blog post that explores Valkey's new 9.2 commands as well as a couple of other options developers should know, with practical examples of how to use most of them in their deployment.

It covers: XDELEX, XACKDEL, MOVE with REPLACE, MULTIIF, ZRANGE XX, SET IFNE, SISMEMBER XX.

It also provides examples, an explanation for each command and links to their respective documentation equivalent if available.

**NOTE:** Some of these commands are still not finalized or merged.

### Check List


content/blog/2026-09-15-valkey-9.2-new-commands/index.md
```diff
@@ -0,0 +1,250 @@
...
+    (integer) -1
+    ```
+
+In this example, `SISMEMBER nosuchkey a XX` returns `-1` because the key `nosuchkey` doesn't exist. Without XX, that same call would just return `0`, indistinguishable from a real "not a member" result. For more information, see the [SISMEMBER command](https://valkey.io/commands/sismember/).
+
+## Try it yourself
+
+Most of what's covered here is still taking shape some of these commands are proposals working through review, and syntax may shift before they land. But that's exactly why now is a good time to get familiar with the direction Valkey is heading: fewer round trips for transactional workflows, and replies that finally let you tell "empty" apart from "doesn't exist" without extra calls.
+
+If you're still running an older version of Valkey, this is a good moment to start planning your upgrade path as each release since has closed real gaps like these, and 9.2 looks to continue that trend. In the meantime, you don't have to wait for a tagged release to see this code in action, everything shown here was tested by building Valkey directly from its `unstable` branch, and you can do the same.
```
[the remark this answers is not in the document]
Andriciuc: Indeed, the CTA will be worked on. Thank you.

content/blog/2026-09-15-valkey-9.2-new-commands/index.md
```diff
@@ -0,0 +1,250 @@
++++
+title = "What’s New in Valkey 9.2: Commands and Options Developers Should Know"
+date = 2026-09-15
+description = "Exploring Valkey's new 9.2 commands as well as options developers should know, with practical examples of how to use most of them in your deployment." 
+authors =  ["dragosandriciuc"]
+[taxonomies]
+blog_type = ["Community Highlight"]
++++
+
+<!-- Some Valkey 9.2 features covered here are still under development and their syntax or behavior may change before release. In particular, feedback is welcome on the sections covering MULTIIF, XACKDEL, and other recently proposed command changes -->
```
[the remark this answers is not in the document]
Andriciuc: Ack, pending resolution based on PR authors reviews.

content/blog/2026-09-15-valkey-9.2-new-commands/index.md
```diff
@@ -0,0 +1,250 @@
...
+## `MULTIIF` for optimistic-locking workflows
+
+<!-- The feature is still being finalized: it started life as a dedicated MULTIIF command, but recent core-team discussion is leaning toward folding the conditions into EXEC itself (e.g. EXEC IFEQ key value, EXEC NX key, EXEC XX key). Ask about it's functionality once merged.
+-->
+
+Starting off with one of the newest proposed commands added in the 9.2 update, `MULTIIF`. This command allows you to attach transaction preconditions before `EXEC`, avoiding the extra `WATCH` round trip that trips up pipelined clients.
+
+Consider the following example:
+
+    ```bash
```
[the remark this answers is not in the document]
Andriciuc: Ack, will fix and push.

Andriciuc:
Hello @nickiaq, I have made a blog post explaining XDELEX and XACKDEL from [#3467](https://github.com/valkey-io/valkey/pull/3467) and [#3466](https://github.com/valkey-io/valkey/pull/3466) respectively, could you please take a look and confirm their functionality as described here? I saw they're still not merged in unstable. Thank you!

Andriciuc:
Hi @bandalgomsu, I have written a blog about the [MULTIIF ](https://github.com/valkey-io/valkey/pull/4019) command. I know it’s currently being refactored, so I wanted to check whether the way I’ve described its behavior is accurate for the current implementation.

If it is, I’ll keep an eye on the refactoring and update the blog as needed before the PR is merged. Do you also have an idea of when you expect the command/API to be stable? Thank you!

content/blog/2026-09-15-valkey-9.2-new-commands/index.md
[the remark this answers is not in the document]
Andriciuc: Removed indentation however left the bash fence mention. Will push soon.

content/blog/2026-09-15-valkey-9.2-new-commands/index.md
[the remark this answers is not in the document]
Andriciuc: I decided to set the first paragraph as a comment, my reasoning is that if the above commands will be merged, I can safely restructure the paragraph without having reviewers be confused as to why it currently says the commands are taking shape.

madolson (commented):
Broadly speaking, I don't like these small little one-off command blogs. We'll basically have one launch launch blog that will propagate up the important commands. Like https://valkey.io/blog/valkey-9-1-delivers-improvements-in-security-performance-and-more/. I don't think we should have a dedicated blog which really is just documentation.

Andriciuc:
> Broadly speaking, I don't like these small little one-off command blogs. We'll basically have one launch launch blog that will propagate up the important commands. Like https://valkey.io/blog/valkey-9-1-delivers-improvements-in-security-performance-and-more/. I don't think we should have a dedicated blog which really is just documentation.

@madolson Hi Madelyn! Thanks for the feedback, let me offer a slightly different angle. Several of the commands covered here (MULTIIF, SET IFNE, XACKDEL, XDELEX) are still unmerged PRs. They don't have a home on valkey.io's docs yet, since docs only cover shipped commands so this isn't duplicating documentation at the moment, it's covering ground docs can't cover until these land. I also built each of these from source and tested them against a live server (including the actual race-condition/fan-out behavior as much as I understood, not just the happy path), which is a level of hands-on verification a docs page or release-notes bullet doesn't really do.

I also think there's a real discoverability gap this fills. As a new user, going from "I want to know what's new" to the actual GitHub PRs and release-plan issues takes several hops, and a lot of it is scattered across threads with contested naming and unmerged syntax. A single post that gathers the promising ones with tested examples closes that gap.

That said, I completely understand the point about this reading too close to documentation. Would it make sense to keep the flagship 9.2 launch post as the primary announcement (like the 9.1 one you linked), and treat posts like this as a secondary "Community Highlight" track? They can be aimed at readers who want the hands-on detail and status on individual proposals (especially the ones still in review). Happy to adjust framing/scope if that's a better fit.

I am also still waiting for replies from the PR owners in regards to them merging their work. 🙏

content/blog/2026-09-15-valkey-9.2-new-commands/index.md
```diff
@@ -0,0 +1,250 @@
...
+
+## `XACKDEL` to delete acknowledged messages
+
+<!-- The feature is still being finalized. -->
+
+`XACKDEL` is a stream command that acknowledges one or more messages and (conditionally) deletes them from the stream.
+
+**Note:** This is particularly useful when multiple consumer groups independently process the same stream and you need to reclaim entries without deleting messages that another group still needs.
+
+It works by periodically reclaiming space once messages are acknowledged by all groups. You could use a Lua script that replicates the `XACKDEL` logic, but `XACKDEL` together with `ACKED` mode makes Lua script unnecessary.
```
nickiaq: > It works by periodically reclaiming space ...

This might give the impression that `XACKDEL` has some async/background component server-side, where instead it's that the message is deleted only when the last consumer group on the stream acknowledges with `XACKDEL`.

content/blog/2026-09-15-valkey-9.2-new-commands/index.md
```diff
@@ -0,0 +1,250 @@
...
+
+`XDELEX` is an extension of the Valkey Streams [`XDEL` command](https://valkey.io/commands/xdel/) that allows you to delete one or more stream messages with more control over how those message entries are deleted concerning consumer groups.
+
+The command supports three deletion modes:
+
+- `KEEPREF` (default): Deletes the stream entry but leaves PEL references intact in all consumer groups
+- `DELREF`: Deletes the stream entry and forcibly removes it from all consumer group PELs
+- `ACKED`: Only deletes the entry once every consumer group has acknowledged or passed it — safe for fan-out topologies
+
+**Note:** The command returns a per-ID integer array: 1 for deleted, 2 for exists-but-not-yet-deletable (ACKED mode only), and -1 for not found.
```
nickiaq: The `-1` response code also covers when the message was never delivered to the group or already ack'ed by it.

content/blog/2026-09-15-valkey-9.2-new-commands/index.md
```diff
@@ -0,0 +1,250 @@
...
+`XACKDEL` is a stream command that acknowledges one or more messages and (conditionally) deletes them from the stream.
+
+**Note:** This is particularly useful when multiple consumer groups independently process the same stream and you need to reclaim entries without deleting messages that another group still needs.
+
+It works by periodically reclaiming space once messages are acknowledged by all groups. You could use a Lua script that replicates the `XACKDEL` logic, but `XACKDEL` together with `ACKED` mode makes Lua script unnecessary.
+
+The command supports the following deletion modes:
+
+- `KEEPREF` (default, implicit): acknowledges and deletes messages immediately, leaving `PEL` references in other groups
+- `ACKED`: deletes messages only once every consumer group has acknowledged or passed the message, making it safe for fan-out stream topologies, if a message was acknowledged but not deleted then the response is 2
```
nickiaq: "Acknowledged or passed" isn't quite accurate because both conditions are neccessary.

In code, the conditions are stated in the negative: ACKED requires that the message is not in any group's PEL (each group either acknowledged it or never had it delivered) and that the message is at or below every group's last delivered ID (so no group can still deliver it later). A delivered but unacknowledged message stays in the PEL even after the group's last delivered ID has passed it, and `XGROUP SETID` / `XCLAIM ... LASTID` can move the last delivered ID without touching the PEL, so last delivered ID alone never implies "not pending."

Not sure how to condense that down into something more friendly for a blog post. Maybe: "ACKED deletes a message only once no consumer group still needs it: none has it pending (each has either acknowledged it or never picked it up), and no group can still deliver it later."

---

Also, the order of the options is `KEEPREF`, `DELREF`, `ACKED` in case we want that to be consistent in the 2 sections here and with the ordering defined in the command typeahead/autocomplete on the CLI.

content/blog/2026-09-15-valkey-9.2-new-commands/index.md
```diff
@@ -0,0 +1,250 @@
...
+
+<!-- The feature is still being finalized. -->
+
+`XDELEX` is an extension of the Valkey Streams [`XDEL` command](https://valkey.io/commands/xdel/) that allows you to delete one or more stream messages with more control over how those message entries are deleted concerning consumer groups.
+
+The command supports three deletion modes:
+
+- `KEEPREF` (default): Deletes the stream entry but leaves PEL references intact in all consumer groups
+- `DELREF`: Deletes the stream entry and forcibly removes it from all consumer group PELs
+- `ACKED`: Only deletes the entry once every consumer group has acknowledged or passed it — safe for fan-out topologies
```
nickiaq: Same caveats apply here around "acknowledged or passed".

nickiaq (commented):
Hey @Andriciuc! Thanks for putting this together. Not sure where this is going given @madolson's comments. But, I've added my comments below.

Also, a few questions on timing & docs:

1. Pending round 2 of review, it sounds like things are lining up for `XACKDEL` and `XDELEX` to be included in 9.2; if so, should I push an adjustment to the PR's to change `"since": "10.0.0"` in the command JSON to `"since": "9.2.0"`?
2. For docs, is the norm to send a PR(s) documenting `XACKDEL` and `XDELEX` to [valkey-io/valkey-doc](https://github.com/valkey-io/valkey-doc)?

bandalgomsu:
Hi @Andriciuc!

I have updated the proposal to use regular MULTI with conditional EXEC arguments:
```
MULTI
...
EXEC [IFEQ <key> <value> | IFNE <key> <value> | NX <key> | XX <key>]...
```

The conditions are evaluated at EXEC time. Please refer to the PR for the latest details. Thanks!

content/blog/2026-09-15-valkey-9.2-new-commands/index.md
```diff
@@ -0,0 +1,250 @@
++++
+title = "What’s New in Valkey 9.2: Commands and Options Developers Should Know"
+date = 2026-09-15
+description = "Exploring Valkey's new 9.2 commands as well as options developers should know, with practical examples of how to use most of them in your deployment." 
+authors =  ["dragosandriciuc"]
+[taxonomies]
+blog_type = ["Community Highlight"]
```
stockholmux: Not really the right blog type - these are more for things about the community.

stockholmux (changes requested):
I think this blog post needs a bit of a nudge in another direction. The general bias of our new version coverage tends to lean toward operational aspects of Valkey. So, where I think this should go is directly addressing new items in 9.2 that are useful not to people who run Valkey but those who write software that _uses_ Valkey. Also, at time of review, this seems more like documentation than a blog post.

What I'd like to see is an indication of why these commands are important or useful. Documentation is evergreen a describes what something does where as blog posts are slice-in-time, relevant for time of publishing. We should shift it toward the latter.

So, in other words, from the perspective of a developer: what can you do that wasn't easy or possible in 9.1 and is possible/easy in 9.2.

content/blog/2026-09-15-valkey-9.2-new-commands/index.md
```diff
@@ -0,0 +1,250 @@
...
+(integer) 3
+127.0.0.1:6379> sismember myset a
+(integer) 1
+127.0.0.1:6379> sismember myset z
+(integer) 0
+127.0.0.1:6379> sismember myset z XX
+(integer) 0
+127.0.0.1:6379> sismember nosuchkey a
+(integer) 0
+127.0.0.1:6379> sismember nosuchkey a XX
```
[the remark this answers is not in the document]
Andriciuc: Done, will push soon.

content/blog/2026-09-15-valkey-9.2-new-commands/index.md
Andriciuc: Set to: Technical Deep Dive, as we're offering solutions for specific problems with examples as well. I reorganized the information in the blog using the following format: Developer problem -> what was awkward in 9.1 -> what 9.2 changes -> why you should care / example.

This reorganizes the wording quite a bit to make it less technical and to simplify the stack. Thank you for guiding this on the right track! Updating as we speak.

I also believe that the original descriptions will be extremely useful to be added in our documentation as @madolson has pointed, thank you again for the feedback on the original post Madelyn.

Andriciuc:
> Hey @Andriciuc! Thanks for putting this together. Not sure where this is going given @madolson's comments. But, I've added my comments below.
> 
> Also, a few questions on timing & docs:
> 
> 1. Pending round 2 of review, it sounds like things are lining up for `XACKDEL` and `XDELEX` to be included in 9.2; if so, should I push an adjustment to the PR's to change `"since": "10.0.0"` in the command JSON to `"since": "9.2.0"`?
> 2. For docs, is the norm to send a PR(s) documenting `XACKDEL` and `XDELEX` to [valkey-io/valkey-doc](https://github.com/valkey-io/valkey-doc)?

Hi @nickiaq ! Thank you for the review! I addressed all of your feedback and will push an update soon. Also let me answer your questions:

1. Since 9.2.0 is great if you absolutely have confirmed that,  
2. For docs, you can create an issue in the Issues category, and someone will pick it up and create a PR for it, but since I've already worked on it I will make the PR for you for both the new commands, if you'd like please go ahead and make the issue though just to "get into the groove" of how it works usually 

Thank you again 🙏

Andriciuc:
> Hi @Andriciuc!
> 
> I have updated the proposal to use regular MULTI with conditional EXEC arguments:
> 
> ```
> MULTI
> ...
> EXEC [IFEQ <key> <value> | IFNE <key> <value> | NX <key> | XX <key>]...
> ```
> 
> The conditions are evaluated at EXEC time. Please refer to the PR for the latest details. Thanks!

Perfect, great to know! I have updated the description and content properly, thank you too @bandalgomsu !

content/blog/2026-09-15-valkey-9.2-new-commands/index.md
Andriciuc: Fixed, updated the paragraph to: With `ACKED` mode, the message is deleted only when the last consumer group processes it with `XACKDEL`.

Andriciuc: Small note, updated to: With `ACKED` mode, the message is deleted only when the last consumer group processes it with `XACKDEL`,meaning none has it pending (each has either acknowledged it or never picked it up), and no group can still deliver it later.

Wanted to ensure the behavior is strictly explained.

content/blog/2026-09-15-valkey-9.2-new-commands/index.md
Andriciuc: I used that phrase to update the ACKED message: `ACKED`: deletes a message only once no consumer group still needs it, meaning none has it pending (each has either acknowledged it or never picked it up), and no group can still deliver it later

It should not force readers to go into PEL mechanics, as it's easy to understand as is, thank you!

content/blog/2026-09-15-valkey-9.2-new-commands/index.md
Andriciuc: Oh! Updated: `-1` when the message wasn't found, wasn't delivered to the group, or was already acknowledged by it.

Thank you for confirming that!

content/blog/2026-09-15-valkey-9.2-new-commands/index.md
Andriciuc: Updated with the above: Deletes a message only once no consumer group still needs it, meaning none has it pending (each has either acknowledged it or never picked it up), and no group can still deliver it later.

madolson:
> So, where I think this should go is directly addressing new items in 9.2 that are useful not to people who run Valkey but those who write software that uses Valkey. Also, at time of review, this seems more like documentation than a blog post.

Yeah, I think Kyle identified my concern better than I did. As this reads, it's not really a technical deep dive, it's more of announcement. So there are six'ish new commands, and here is where I would place them:
`ZRANGE XX`, `SISMEMBER XX`, `MOVE [REPLACE]` are just simplifying multi-execs into single commands today. Maybe there is a story of simplifying here, but this doesn't seem blog worthy.

`EXEC IFEQ` is just an alternative to Lua. The fact that there is zero mention of Lua in the blog you're missing the main alternative. People aren't replacing watch with `EXEC`, they might be replacing Lua with it. 

`SET IFNE` is structurally the same, but has a very targeted locking use case, on brand for us but also not really that interesting. It also allows allows suppression of idempotent writes, which is neat, imo. 

`XDELEX`/`XACKDEL` is more novel I guess.  I think this is the only one that isn't trivial to do with Lua, you would need to do a lot of parsing. 

A genuinely more interesting blog to me is "Why you shouldn't use Lua for everything". And highlight how commands innovate.

content/blog/2026-09-15-valkey-9.2-new-commands/index.md
```diff
@@ -0,0 +1,275 @@
...
+
+## `SISMEMBER XX` for distinguishing missing keys from non-members
+
+Sometimes your application needs to distinguish between "the user isn't in this set" and "the set doesn't exist."
+
+Before 9.2, `SISMEMBER` couldn't distinguish the two cases by itself. For example, if you ran `SISMEMBER users:online alice` you'd get `0`. But what does `0` actually mean? That Alice isn't online? That `users:online` don't exist?
+
+There is no way to tell. You could use `EXISTS users:online` then `SISMEMBER users:online alice` but now you've got another round trip operation, and your application needs to combine two pieces of information together.
+
+With Valkey 9.2, `SISMEMBER` now has an `XX` option that makes that distinction explicit:
```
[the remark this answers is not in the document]
Andriciuc: This is factually wrong, see the following PR: https://github.com/valkey-io/valkey/pull/2978

Ignore this error in the future please.

content/blog/2026-09-15-valkey-9.2-new-commands/index.md
```diff
@@ -0,0 +1,275 @@
...
+
+`XDELEX` is an extension of the Valkey Streams [`XDEL` command](https://valkey.io/commands/xdel/) that allows you to delete one or more stream messages with more control over how those message entries are deleted concerning consumer groups.
+
+The command supports three deletion modes:
+
+- `KEEPREF` (default): deletes the stream entry but leaves PEL references intact in all consumer groups
+- `DELREF`: deletes the stream entry and forcibly removes it from all consumer group PELs
+- `ACKED`: deletes a message only once no consumer group still needs it, meaning none has it pending (each has either acknowledged it or never picked it up), and no group can still deliver it later
+
+**Note:** The command returns a per-ID integer array: `1` for deleted, `2` for exists-but-not-yet-deletable (`ACKED` mode only), and `-1` when the message wasn't found, wasn't delivered to the group, or was already acknowledged by it.
```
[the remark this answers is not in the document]
Andriciuc: https://github.com/valkey-io/valkey/pull/3466 and https://github.com/valkey-io/valkey/pull/3467 handle that, and will be either mentioned as references or retroactively added into the blog once documentation for them is made.

content/blog/2026-09-15-valkey-9.2-new-commands/index.md
```diff
@@ -0,0 +1,275 @@
...
+
+`XACKDEL` is a stream command that acknowledges one or more messages and (conditionally) deletes them from the stream.
+
+This is particularly useful when multiple consumer groups independently process the same stream and you need to reclaim entries without deleting messages that another group still needs.
+
+With `ACKED` mode, the message is deleted only once no consumer group still needs it, meaning none has it pending (each has either acknowledged it or never picked it up), and no group can still deliver it later.
+
+The command supports the following deletion modes:
+
+- `KEEPREF` (default, implicit): acknowledges and deletes messages immediately, leaving `PEL` references in other groups
```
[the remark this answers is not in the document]
Andriciuc: I will ensure this is done on the next update, thank you.

content/blog/2026-09-15-valkey-9.2-new-commands/index.md
```diff
@@ -0,0 +1,352 @@
...
+[taxonomies]
+blog_type = ["Technical Deep Dive"]
++++
+
+A lot of application logic built on top of Valkey ends up in one of two places: either scattered across multiple round trips in your client code, or pushed into a Lua script so the server can do it atomically in one call.
+Both work, but they have different costs: extra round trips add latency and race windows, while scripts add something to write, ship, cache, and get right.
+
+Several of Valkey 9.2's changes exist specifically to shrink that second category: giving plain commands the conditional logic that used to mean either extra round trips or a script.
+
+**Note:** If you want the background on how Lua scripting works in Valkey before diving in, the [Introduction to Eval Scripts](https://valkey.io/topics/eval-intro/) docs page is a good starting point.
```
stockholmux: I don't think "Note" asides are very nice to read in this format. Just integrate this sentence as part of the previous paragraph without the note.

content/blog/2026-09-15-valkey-9.2-new-commands/index.md
```diff
@@ -0,0 +1,352 @@
...
+
+Let's look at where 9.2 actually changes things.
+
+## `SET IFNE` to set a key when its value doesn't match a specified value
+
+Sometimes your application wants to update a value only if it hasn't already been set to something specific, to replace a stale default or placeholder, for example, without touching it if another process has already moved it on.
+
+Before 9.2, doing this safely meant a `GET` first, checking the value in your application code, then conditionally issuing a `SET` which is an extra round trip, and a race window between the `GET` and the `SET` where another client could write in between.
+A Lua script can close that race window in one round trip. However, it's still a script you have to write for a pattern that comes up constantly.
+However, it's still a script you have to write for a pattern that comes up constantly.
```
stockholmux: Clankers are right - you've got a dupe. Please delete.

content/blog/2026-09-15-valkey-9.2-new-commands/index.md
```diff
@@ -0,0 +1,352 @@
...
+127.0.0.1:6599> SET mykey{foo}2 222
+127.0.0.1:6599> EXEC
+```
+
+Or you could write a Lua script and let the server do the check and the writes in a single round trip:
+
+```lua
+-- KEYS[1] = version key, ARGV[1] = expected value, ARGV[2] = new value
+-- KEYS[2], KEYS[3] = the other keys, ARGV[3], ARGV[4] = their new values
+local current = redis.call('GET', KEYS[1])
```
stockholmux: `redis.` still works, but please change to `server.` in all cases.

content/blog/2026-09-15-valkey-9.2-new-commands/index.md
```diff
@@ -0,0 +1,352 @@
...
+127.0.0.1:6899> XLEN s4
+(integer) 0
+```
+
+**Note:** Your `XADD` will return a different ID, ensure you substitute it throughout.
+
+From the above example you can see that the first `XDELEX ... ACKED` call returns `2`, because `grp` still has the message pending, and `XLEN` confirms it's still in the stream.
+Once `XACK` explicitly acknowledges it for `grp`, the same `XDELEX ... ACKED` call returns `1` and `XLEN` drops to `0` which means the message is only actually removed once **no** consumer group still needs it.
+
+Lua isn't going away, and it shouldn't. When an operation needs application-specific logic that Valkey doesn't expose as a command, scripting is still a useful way to keep that logic atomic. The point of these 9.2 changes is to make several common patterns no longer need a script at all.
```
stockholmux: This feels like a conclusion paragraph but it's in the middle...

content/blog/2026-09-15-valkey-9.2-new-commands/index.md
```diff
@@ -0,0 +1,352 @@
...
+
+## Sharper replies, one fewer round trip: `ZRANGE XX`, `SISMEMBER XX`, and `MOVE ... REPLACE`
+
+The rest of 9.2's changes covered here follow a simpler pattern: an existing command gains an option that removes a companion call your application used to need, an `EXISTS` check, or a manual copy-then-delete.
+None of these need Lua to work around; they're about collapsing two round trips into one.
```
stockholmux: Okay - I don't really agree with this statement. The round trips also introduce a concurrency problem so people would use Lua in these to ensure that the evaluation logic isn't interrupted.

stockholmux: Reformat the other new commands with their equal Lua to illustrate 9.2's Lua avoidance. 

You could also combine Lua example for the two XX cases since they are functionally the same "If EXISTS, then one return value, if not, result of command value"

content/blog/2026-09-15-valkey-9.2-new-commands/index.md
```diff
@@ -0,0 +1,352 @@
...
+(integer) -1
+```
+
+Now your application can distinguish:
+
+- `1` means Alice is a member
+- `0` means the set exists, but the member isn't in it
+- `-1` means the set doesn't exist
+
+### `MOVE ... REPLACE`
```
stockholmux, line 392: The pattern that MOVE...REPLACE enables can be done with a MULTI/EXEC but you leave the database number mutated in the client, so you have to switch it back atomically (if you don't you'll be writing data on the wrong database. The safest way to do this would be with a Lua script.

Something like:
```
EVAL "server.call('SELECT', ARGV[1])\r\nserver.call('DEL', KEYS[1])\r\nserver.call('SELECT', ARGV[2])\r\nserver.call('MOVE', KEYS[1],ARGV[1])\r\nserver.call('SELECT', ARGV[1])" 1 foo 0 5
```
 
This mess would overwrite key `foo` on database 0 with key `foo` from database 5.

Alternately, you could do some sort of unholy `RENAME` to accomplish the same thing but you'd need key tags to make it work.

Regardless, MOVE ... REPLACE is loads cleaner.

Details: https://github.com/valkey-io/valkey/issues/2278

content/blog/2026-09-15-valkey-9.2-new-commands/index.md
```diff
@@ -0,0 +1,352 @@
...
+
+For more information, see the [MOVE command documentation](https://valkey.io/commands/move/).
+
+## Try it yourself
+
+Taken together, these changes point in one direction: less application logic living in round trips or in Lua, more of it expressed directly in the commands themselves.
+Whether you're locking on a version key, cleaning up a fan-out stream, or just tired of an extra `EXISTS` call, 9.2 gives you a way to say what you mean in one call instead of building it yourself.
+
+If you're running an older version of Valkey, these examples are a good way to identify application logic that can become simpler after upgrading.
+[Clone the repository](https://github.com/valkey-io/valkey), check out the relevant branch or commit, build Valkey with `make`, and try the commands for yourself.
```
stockholmux: This is going to go out post 9.2, so we can say something like 'get latest version from the downloads page'.

stockholmux (changes requested):
Let's reframe everything to Lua.

content/blog/2026-09-15-valkey-9.2-new-commands/index.md
Andriciuc: Integrated, thank you!

content/blog/2026-09-15-valkey-9.2-new-commands/index.md
Andriciuc: Deleted, missed while rewording the paragraphs T_T, thank you!

content/blog/2026-09-15-valkey-9.2-new-commands/index.md
Andriciuc: Oof, thank you! Fixed, usually I don't rename commands so this was a great obs.

content/blog/2026-09-15-valkey-9.2-new-commands/index.md
Andriciuc: Updated! Also, I left the link to the github discussions, I feel like that would help direct more traffic to the repo by just being visible here.

content/blog/2026-09-15-valkey-9.2-new-commands/index.md
Andriciuc: Moved in the conclusions, thanks for catching it 🙏

content/blog/2026-09-15-valkey-9.2-new-commands/index.md
Andriciuc: Perfect feedback, thank you! I hadn't taken into account that two separate calls can be interleaved by another client, I'm checking to see how I can add the Lua mentions based on the previous examples too, keeping it simple.

content/blog/2026-09-15-valkey-9.2-new-commands/index.md
Andriciuc: Thank you! I observed one difference with the Lua script example, there is a SELECT leak that appears to be specific to MULTI/EXEC rather than Lua. 

If I understood correctly, SELECT 1 persists inside the block (when a transaction occurs) on the connection after EXEC, so the next SET ends up in db1. With a script, it restores the connection's selected database when the script exits, so the script still works without switching back to the source towards the end. I ended up writing the section around that distinction using MULTI/EXEC as the footgun, Lua as a working but fiddly workaround, and MOVE ... REPLACE as the much simpler option. I also added the link to #2278 if readers want to see the actual implementation of the feature!

content/blog/2026-09-15-valkey-9.2-new-commands/index.md
Andriciuc: Added equal Lua's to illustrate avoidance.

content/blog/2026-09-15-valkey-9.2-new-commands/index.md
```diff
@@ -0,0 +1,472 @@
...
+
+A lot of application logic built on top of Valkey ends up in one of two places: either scattered across multiple round trips in your client code, or pushed into a Lua script so the server can do it atomically in one call.
+Both work, but they have different costs: extra round trips add latency and race windows, while scripts add something to write, ship, cache, and get right.
+
+Several of Valkey 9.2's changes exist specifically to shrink that second category: giving plain commands the conditional logic that used to mean either extra round trips or a script.
+
+If you want the background on how Lua scripting works in Valkey before diving in, the [Introduction to Eval Scripts](https://valkey.io/topics/eval-intro/) documentation page is a good starting point.
+
+Let's look at where 9.2 actually changes things.
+
```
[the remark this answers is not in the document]
Andriciuc: The PR for XACKDEL and XDELEX is https://github.com/valkey-io/valkey/pull/4629 and it has been merged. Keeping them in.

content/blog/2026-09-15-valkey-9.2-new-commands/index.md
```diff
@@ -0,0 +1,472 @@
...
+A Lua script avoids that, because Valkey restores the connection's selected database when the script exits:
+
+```lua
+-- KEYS[1] = the key, ARGV[1] = source db, ARGV[2] = destination db
+server.call('SELECT', ARGV[2])
+server.call('DEL', KEYS[1])
+server.call('SELECT', ARGV[1])
+server.call('MOVE', KEYS[1], ARGV[2])
+return 1
```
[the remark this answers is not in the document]
Andriciuc: db0 ends empty, db1 ends with "active" which is the value that was in db0 before the script ran, it is exactly the intended MOVE ... REPLACE outcome

PR #662, opened 2026-09-05
Rework the homepage and download pages around release versions

madolson:
This PR mostly attempts to tweak the main page so that there is less dead space and makes new versions cleaner. Added some text about the major new features + some links.

Performance got dropped down to one of the lower tiles. 

The blog cards lost their thumbnail. Every featured image is 1920x1080 and the 230px slot cropped them to roughly square, so each card showed an arbitrary slice of a gradient. Restoring the full 16:9 image is not an option either: at the column's ~620px width that is a 350px banner per card. The cards are text now, matching the release cards opposite, and a fourth post fills the height the images used to take. The blog index link moved up beside the heading.

See below for images
<details>

image

Valkey homepage after the rework
</details>

### DIFF
--- a/content/_index.md
+++ b/content/_index.md
@@ -6,6 +6,9 @@ repos_heading= "Featured repositories"
 github_org= "https://github.com/valkey-io"
 github_org_text= "All repos in the Valkey GitHub organization"
 
+# Number of blog posts shown in the "Latest Blog Posts" panel.
+latest_blog_posts= 3
+
 headline= "Valkey: an open source, in-memory data store"
 long_description= ""
 
@@ -14,12 +17,6 @@ headings = ["FAST. RELIABLE.", "OPEN SOURCE, FOREVER."]
 subtitle = "Valkey is an open source (BSD) high-performance key/value datastore that supports a variety of workloads such as caching, message queues, and can act as a primary database. The project is backed by the Linux Foundation, ensuring it will remain open source forever."
 button_text = "GET STARTED"
 button_url = "/topics/quickstart"
-
-[[extra.documentation_cards]]
-title = "Install"
-description = "Step-by-step instructions on how to install and configure Valkey for first-time users."
-button_text = "See Installation Guide"
-button_url = "/topics/installation"
 
 [[extra.documentation_cards]]
 title = "Usage guide"
@@ -35,8 +32,7 @@ button_url = "/commands"
 
 [[extra.documentation_cards]]
 title = "Clients"
-description = "Official Valkey client libraries include support for:"
-features = ["Python", "Java", "Go", "Node.js", "PHP", "C#", "Ruby"]
+description = "Official Valkey client libraries for Python, Java, Go, Node.js, PHP, C#, Ruby, and more."
 button_text = "Learn More"
 button_url = "/clients"
 
--- a/content/_index.md
+++ b/content/_index.md
@@ -7,7 +7,7 @@ github_org= "https://github.com/valkey-io"
 github_org_text= "All repos in the Valkey GitHub organization"
 
 # Number of blog posts shown in the "Latest Blog Posts" panel.
-latest_blog_posts= 3
+latest_blog_posts= 4
 
 headline= "Valkey: an open source, in-memory data store"
 long_description= ""
--- a/content/_index.md
+++ b/content/_index.md
@@ -36,6 +36,12 @@ description = "Official Valkey client libraries for Python, Java, Go, Node.js, P
 button_text = "Learn More"
 button_url = "/clients"
 
+[[extra.documentation_cards]]
+title = "Performance"
+description = "Throughput trends across every Valkey version."
+button_text = "Explore"
+button_url = "/performance/"
+
 [[extra.download_ctas]]
 text= "Get Valkey"
 url=  "/download/"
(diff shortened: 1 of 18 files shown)


templates/index.html
```diff
@@ -122,21 +122,24 @@ <h3>{{sidebar_category.title}}</h3>
...
-                        <li>{{ feature }}</li>
-                        {% endfor %}
-                    </ul>
-                    {% endif %}
                     <a class="btn" href="{{ card.button_url }}">{{ card.button_text }}</a>
                 </div>
             </div>
             {% endfor %}
         {% endif %}
+        {% set perf = load_data(path="_data/perf.toml") %}
```
stockholmux: I don't think this makes a lot of sense under documentation. I recall quite a bit of consternation about getting performance numbers at the top of the page when we implemented it originally, so _someone_ cares where it goes and I doubt this is satisfactory.

stockholmux (changes requested):
Generally, I like it. I have a few tweaks/modifications I'd like to see:

1) Documentation seems like a super odd fit for perf dashboards.
2) I'm worried a little bit about the release sequence and how this might create a content race condition. Generally, we make sure the version is live on the page (done by the release manager) then push the release blog. Adding the release line bit complicates it, especially considering it needs to link the blog. We should think this through and see if there is ways to make it easily sequential and document who does what.
3) While I like the release lines summaries, how would we handle a condition where there is a super critical update required on a patch? OpenSearch had this a couple of times and they used a similar content box to communicate this - but here we're tying that info to a release line, not a specific patch (each patch version had a description for this reason). Any thoughts on this?

madolson:
> Documentation seems like a super odd fit for perf dashboards.

Yeah. Honestly, do we just drop it from there? I really don't know where to put it but it's current position makes no sense. I ended up moving it down and changing the header section.

> I'm worried a little bit about the release sequence and how this might create a content race condition. Generally, we make sure the version is live on the page (done by the release manager) then push the release blog. Adding the release line bit complicates it, especially considering it needs to link the blog. We should think this through and see if there is ways to make it easily sequential and document who does what.

This will now throw an error during the build, so we still have a sequence, which is we need to generate the new markdown for the release line before the release happens.

> While I like the release lines summaries, how would we handle a condition where there is a super critical update required on a patch? OpenSearch had this a couple of times and they used a similar content box to communicate this - but here we're tying that info to a release line, not a specific patch (each patch version had a description for this reason). Any thoughts on this?

I think it would make sense to do Low, Medium, high, Security as text next to each release. Follow up PR?

content/_index.md
```diff
@@ -35,11 +32,16 @@ button_url = "/commands"
 
 [[extra.documentation_cards]]
 title = "Clients"
-description = "Official Valkey client libraries include support for:"
-features = ["Python", "Java", "Go", "Node.js", "PHP", "C#", "Ruby"]
+description = "Official Valkey client libraries for Python, Java, Go, Node.js, PHP, C#, Ruby, and more."
```
stockholmux, line 35: I know 'official' is not new in this PR, but I think we should align with the terminology on the linked page of 'recommended'. People tend to bat around the term 'official' too much and we've never defined it. The linked page has ones housed on the  valkey-io GitHub (a nominal official line...) but also ones that aren't like redisson.

content/_index.md
```diff
@@ -15,12 +18,6 @@ subtitle = "Valkey is an open source (BSD) high-performance key/value datastore
...
 
-[[extra.documentation_cards]]
-title = "Install"
-description = "Step-by-step instructions on how to install and configure Valkey for first-time users."
-button_text = "See Installation Guide"
-button_url = "/topics/installation"
-
 [[extra.documentation_cards]]
 title = "Usage guide"
 description = "Detailed documentation on the various datatype supported by Valkey and best practices."
```
stockholmux, line 23: Not new to this PR but I did notice that we say "various datatype" should be "datatypes"

stockholmux (changes requested):
Comparing back to back to the current site, I see a couple of odd choices - no huge deal breakers from my end outside of the first item.

1) We lost the "Install" tile in your version (and the word 'install' is missing from the entire page...). I feel like that's something people want to find on a homepage for an OSS project.
2) The link colour on the blog posts tiles is inconsistent. On the latest release tile, the links are blue. On the latest blog post tiles they are black. We should try to keep them consistent, at least when on the same background colour.
3) Similarly, "All blog posts >" should probably a different colour to make it very obviously a link. Can we right justify it too? Alternately, we can put it at the bottom of the tile - which is a more common pattern (but potentially causes some contrast issues due to the gradient.
4) the "Learn More" and "Participants" titles are different sizes, which is just a slight bit annoying
5) The  container that holds blog posts and releases and "Learn More" heading has maybe a little too much space between: There is spacing between the inner and outer container, then a space below the container and a space above the text. I would tighten that up the container a bit.

pnbrown (commented):
I really like the home page change with latest release details. It's a shame that doesn't carry through to the download page. 

A bit sad about how far down the performance dropped but it's not a change I feel super passionate about. The new version looks a lot better.

Download page definitely looks a bit cleaner. Curious about Download Valkey vs Download Latest

I think I agree generally with what Kyle is saying too.

content/_index.md
madolson: oops. Will fix along the way.

madolson:
Install is back as a second hero button, light for Get started, dark for Install Valkey.

Blog card titles now take the same link colour as the release links opposite. The index link is #B7C2F7 and sits right justified on the heading row rather than at the bottom of the tile, to keep it off the gradient.

Headings and panel spacing as described.

9.2.0-rc1 rides along so the new release candidate list on the download page has something to render. Happy to pull it back out if you would rather it land in its own release PR.

Per patch severity labels still read as a follow-up to me.

